PR #21, opened 2016-03-30
Remove global Server* server, capture by reference

chakaz:
It seems to be a hack to allow the lambda to use the server, not sure why you'd want that.
If you want to learn more about lambdas, check out this link: http://shaharmike.com/cpp/lambdas-and-functions/

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -9,7 +9,6 @@ using namespace std;
 using namespace uWS;
 
 int connections = 0;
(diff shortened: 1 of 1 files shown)


alexhultman:
I might have fooled you by including functional, but the callbacks are actually not closures, but "raw" C function pointers. I will change them to std::function sometime soon.

PR #39, opened 2016-04-09
Rewrite the ws interface in ES2015

lpinca:
µWS works only on node 4+ so I think it makes sense to use ES2015.
I did not test the changes so there might be bugs.

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -1,98 +1,131 @@
-/* ws interface for Node.js developers */
+'use strict';
 
-const NativeServer = require('./uws_' + process.platform + '_' + process.versions.modules).Server;
 const EventEmitter = require('events');
-const util = require('util');
 
-function Socket(nativeSocket, server) {
-    EventEmitter.call(this);
-    var self = this;
-    this.nativeSocket = nativeSocket;
-    this.server = server;
+const uws = require(`./uws_${process.platform}_${process.versions.modules}`);
+const NativeServer = uws.Server;
 
-    this.send = function(message, options) {
-        if (self.nativeSocket !== null) {
-            var binary = false;
-            if (options !== undefined) {
-                binary = options.binary;
-            }
-            self.server.nativeServer.send(self.nativeSocket, message, binary);
-        } else {
-            /* ignore sends on closed sockets */
-        }
+class Socket extends EventEmitter {
+    /**
+     * Creates a Socket instance.
+     *
+     * @param {Object} nativeSocket Socket instance
+     * @param {Server} server Server instance
+     */
+    constructor(nativeSocket, server) {
+        super();
+        this.nativeSocket = nativeSocket;
+        this.server = server;
     }
 
-    this.close = function() {
-        if (self.nativeSocket !== null) {
-            self.server.nativeServer.close(self.nativeSocket);
-            self.nativeSocket = null;
-        } else {
-            /* ignore close on closed sockets */
-        }
+    /**
+     * Sends a message.
+     *
+     * @param {String|Buffer} message The message to send
+     * @param {Object} options Send options
+     * @public
+     */
+    send(message, options) {
+        /* ignore sends on closed sockets */
+        if (!this.nativeSocket) return;
+
+        const binary = options && options.binary || false;
+        this.server.nativeServer.send(this.nativeSocket, message, binary);
+    }
+
+    /**
+     * Closes the socket.
+     *
+     * @public
+     */
+    close() {
+        /* ignore close on closed sockets */
+        if (!this.nativeSocket) return;
+
+        this.server.nativeServer.close(this.nativeSocket);
+        this.server = this.nativeSocket = undefined;
     }
 }
 
-module.exports.Server = function Server(options) {
-    EventEmitter.call(this);
+class Server extends EventEmitter {
+    /**
+     * Creates a Server instance.
+     *
+     * @param {Object} options Configuration options
+     */
+    constructor(options) {
+        super();
+        /* we currently do only support port as option */
+        this.nativeServer = new NativeServer(options.port);
 
-    /* we currently do only support port as option */
-    this.nativeServer = new NativeServer(options.port);
-    var self = this;
+        this.nativeServer.onConnection((nativeSocket) => {
+            const socket = new Socket(nativeSocket, this);
+            this.nativeServer.setData(nativeSocket, socket);
+            this.emit('connection', socket);
+        });
 
-    this.close = function () {
-        self.nativeServer.close();
+        this.nativeServer.onDisconnection((nativeSocket) => {
+            const socket = this.nativeServer.getData(nativeSocket);
+            socket.emit('close');
+            /* make sure to clear any set data */
+            this.nativeServer.setData(nativeSocket);
+        });
+
+        this.nativeServer.onMessage((nativeSocket, message, binary) => {
+            const socket = this.nativeServer.getData(nativeSocket);
+            socket.emit('message', binary ? message : message.toString());
+        });
     }
-    this.broadcast = function (message, options) {
+
+    /**
+     * Closes the server.
+     *
+     * @public
+     */
+    close() {
+        this.nativeServer.close();
+    }
+
+    /**
+     * Broadcast a message to all sockets.
+     *
+     * @param {String|Buffer} message The message to broadcast
+     * @param {Object} options Broadcast options
+     * @public
+     */
+    broadcast(message, options) {
         /* only listen to binary option */
-        this.nativeServer.broadcast(message, options.binary);
-    };
+        this.nativeServer.broadcast(message, options && options.binary || false);
+    }
 
-    /* original connection callback */
-    this.connectionCallback = function (nativeSocket) {
-        var socket = new Socket(nativeSocket, self);
-        self.nativeServer.setData(nativeSocket, socket);
-        self.emit('connection', socket);
-    };
-
-    /* this function needs to handle the upgrade directly! */
-    this.handleUpgrade = function (request, socket, upgradeHead, callback) {
+    /**
+     * Handles a HTTP Upgrade request.
+     *
+     * @param {http.IncomingMessage} request HTTP request
+     * @param {net.Socket} Socket between server and client
+     * @param {Buffer} upgradeHead The first packet of the upgraded stream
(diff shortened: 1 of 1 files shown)


nodejs/dist/uws.js
```diff
@@ -1,98 +1,134 @@
...
-            self.nativeServer.setData(nativeSocket, socket);
+        nativeServer.onConnection((nativeSocket) => {
+            const sock = new Socket(nativeSocket, this);
+            nativeServer.setData(nativeSocket, sock);
 
             /* internal variables */
-            socket.upgradeReq = request;
-            socket.readyState = socket.OPEN;
-            callback(socket);
+            sock.upgradeReq = request;
```
lpinca: I dropped the `readyState` property on pourpose because `socket.OPEN` here is always `undefined`.

nodejs/dist/uws.js
```diff
@@ -1,98 +1,134 @@
...
+        this.nativeSocket = nativeSocket;
+        this.server = server;
+    }
+
+    /**
+     * Sends a message.
+     *
+     * @param {String|Buffer} message The message to send
+     * @param {Object} options Send options
+     * @return {Boolean} `true` if the message has been sent, else `false`
```
ronkorving: Appreciate the thought behind this boolean. But I would like to keep the door open to an indicator being returned (could be boolean) to expose to the user that buffers are being filled up. This would allow an easy Writable stream wrapper around a Socket, which could be insanely useful in certain use cases (something `ws` *cannot* provide with its API).

nodejs/dist/uws.js
```diff
@@ -1,98 +1,134 @@
...
+        this.nativeServer.onConnection((nativeSocket) => {
+            const socket = new Socket(nativeSocket, this);
+            this.nativeServer.setData(nativeSocket, socket);
+            this.emit('connection', socket);
+        });
 
-    this.close = function () {
-        self.nativeServer.close();
+        this.nativeServer.onDisconnection((nativeSocket) => {
+            const socket = self.nativeServer.getData(nativeSocket);
```
ronkorving: Across this file, please replace `self` with `this`. It does show you didn't test this code once ;)

nodejs/dist/uws.js
```diff
@@ -1,98 +1,134 @@
...
     }
-    this.broadcast = function (message, options) {
+
+    /**
+     * Closes the server.
+     *
+     * @public
+     */
+    close() {
+        this.nativeServer.close();
```
ronkorving: You'll wanna set nativeServer to `undefined` or `null` here.

nodejs/dist/uws.js
```diff
@@ -1,98 +1,134 @@
...
+    handleUpgrade(request, socket, upgradeHead, callback) {
+        const nativeServer = this.nativeServer;
+
         /* register a special connection handler */
-        self.nativeServer.onConnection(function (nativeSocket) {
-            var socket = new Socket(nativeSocket, self);
-            self.nativeServer.setData(nativeSocket, socket);
+        nativeServer.onConnection((nativeSocket) => {
+            const sock = new Socket(nativeSocket, this);
+            nativeServer.setData(nativeSocket, sock);
```
ronkorving: It would seem better to refer to `this.nativeSocket`. There's no need to pull that value from a closure.

nodejs/dist/uws.js
```diff
@@ -1,98 +1,134 @@
...
+     * Handles a HTTP Upgrade request.
+     *
+     * @param {http.IncomingMessage} request HTTP request
+     * @param {net.Socket} Socket between server and client
+     * @param {Buffer} upgradeHead The first packet of the upgraded stream
+     * @param {Function} callback Callback function
+     * @public
+     */
+    handleUpgrade(request, socket, upgradeHead, callback) {
+        const nativeServer = this.nativeServer;
```
ronkorving: In my experience this will not provide performance (possibly the opposite).

lpinca: It actually depends on how many times you access `this.nativeServer`. The cached version is a little faster.
In this case there is no difference in perf and I find it more readable. There is no reason to write `this.nativeServer` every time.

nodejs/dist/uws.js
lpinca: There is no difference in perf.

nodejs/dist/uws.js
lpinca: Yes correct, I mostly only copy pasted this part, I missed it :)
I'm gonna fix it.

lpinca:
Thank you for your review @ronkorving.

alexhultman:
Thanks for the help. I will take a look at this when I get back home.

nodejs/dist/uws.js
ronkorving: I can't tell, but there may be a small difference in memory use since we are now building a closure around our function to hold that nativeServer value. It also prevents us from (if we would ever want to) check if `this.nativeServer` is actually set or not (post-close).

nodejs/dist/uws.js
ronkorving: Interesting that it's faster. In my experience and what I've learned about V8 in these circumstances, is that it's usually better to leave that kind of optimization to V8, since now you may simply be reducing our available CPU registers by one without benefit (this is very evident when people think `for (var i = 0, len = arr.length; i < len; i++) {}` is faster for example than not caching the length, which is simply not the case).

nodejs/dist/uws.js
lpinca: > It also prevents us from (if we would ever want to) check if this.nativeServer is actually set or not (post-close).

Ok, this makes sense.

ronkorving:
Sweet. This looks pretty clean to me now. Some testing would be nice, and due to #41 some changes may end up being in order. The order in which @alexhultman wants to progress with that is up to him of course.

nodejs/dist/uws.js
```diff
@@ -1,98 +1,131 @@
...
+            sock.upgradeReq = request;
+            callback(sock);
 
             /* todo: reset connection handler when the upgrade queue is empty */
             /* this will probably never be noticed as you don't mix upgrade handling */
         });
 
         /* upgrades will be handled immediately */
-        self.nativeServer.upgrade(socket._handle.fd, request.headers['sec-websocket-key']);
+        this.nativeServer.upgrade(socket._handle.fd, request.headers['sec-websocket-key']);
```
lpinca: @alexhultman shouldn't this also take the `upgradeHead`? That buffer is not used anywhere.

nodejs/dist/uws.js
```diff
@@ -1,98 +1,132 @@
...
-    });
-};
-
-util.inherits(Socket, EventEmitter);
-util.inherits(module.exports.Server, EventEmitter);
+    }
+}
+
+exports.Server = Server;
+exports.uws = uws;
```
lpinca: I'm exposing the original interface here for people that don't want to use the `ws` interface.
This might be useful for embedders like Primus where the `Spark` is already wrapping the `nativeSocket`.

lpinca: I would like to avoid doing [this](https://github.com/primus/primus/blob/8f8d111a68b76db2703cba773541a52afb96a567/transformers/uws/server.js#L14-L17).

nodejs/dist/uws.js
alexhultman: Yes but I don't support that yet and it seems to be empty all the time. The Node.js documentation is not very clear about what it is but I guess it holds data received together with the HTTP upgrade request, which is always going to be empty as long as the client is not totally coco and sends both HTTP upgrade request and an WebSocket frame in the same go, without waiting for the response. That is at least how I understand it.

lpinca: > The Node.js documentation is not very clear about what it is but I guess it holds data received together with the HTTP upgrade request

Correct.

alexhultman:
This rewrite reduces the memory consumption from 800 mb to 650 mb for 1 million connections.

alexhultman:
I get this

`Uncaught TypeError: Cannot read property 'getData' of undefined
      at /home/alexhultman/uWebSockets/nodejs/dist/uws.js:68:45`

when I run the primus integration test. It did pass in the old one.

nodejs/dist/uws.js
```diff
@@ -1,98 +1,132 @@
...
+     * Closes the socket.
+     *
+     * @public
+     */
+    close() {
+        /* ignore close on closed sockets */
+        if (!this.nativeSocket) return;
+
+        this.server.nativeServer.close(this.nativeSocket);
+        this.server = this.nativeSocket = undefined;
```
alexhultman: This is causing problems because the close is closing the sockets in the future, so when disconnection callback happens the server needs to be there still.

nodejs/dist/uws.js
```diff
@@ -1,98 +1,132 @@
...
-    this.close = function () {
-        self.nativeServer.close();
+    /**
+     * Closes the server.
+     *
+     * @public
+     */
+    close() {
+        this.nativeServer.close();
+        this.nativeServer = undefined;
```
alexhultman: I meant this line, the close is not closing the sockets this instant, so you cannot set nativeServer to undefined.

nodejs/dist/uws.js
alexhultman: `  var uws = require(path.join(
    path.dirname(require.resolve('uws')),
    'uws_'+ process.platform +'_'+ process.versions.modules
  ));`

This should be just `require('uws')`.

lpinca: Yeah but right now, only the `Server` class is exported so there is no way to get the "native" addon.
I want to use that in primus as I find it odd to use yet another wrapper.

alexhultman: But it seems you are exporting what you need

`exports.Server = Server;`
`exports.uws = uws;`

where

`const uws = require(`./uws_${process.platform}_${process.versions.modules}`);`

Whatever the problem is, just export what you want.

lpinca:
Updated with removal of premature `this.nativeServer` cleanup.

nodejs/dist/uws.js
lpinca: Yes exactly, I added ` exports.uws = uws;` for this specific reason :smile:

alexhultman:
Thanks both for this rewrite. Like I said, it improved memory usage quite a bit.

ronkorving:
Wonderful :)

PR #42, opened 2016-04-10
Windows fixes

peters:
/cc @alexhultman 

Let me know if there is anything you want to change, I'm available on github the next 8 hours :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ WebSocket++ v0.7.0 | µWS is **63x** as lightweight | µWS is **5x** as performa
 
 ## Overview
 
-For Node.js developers, the `ws` interface is exposed:
+For Node.js developers, the `ws` interface is exposed (read [their](https://github.com/websockets/ws/blob/master/doc/ws.md) documentation):
 
 ```javascript
 var WebSocketServer = require('uws').Server /* you replace 'ws' with 'uws' */


src/uWS.cpp
```diff
@@ -17,6 +17,13 @@ using namespace std;
 #endif
 
 #ifdef _WIN32
+#if _MSC_VER >= 1900
+#include <io.h>
+#if !defined(dup)
+#define dup _dup
```
alexhultman: _dup takes an int, we have SOCKET. You have to use WSADuplicateSocket to get similar functionality on Windows.

src/uWS.cpp
```diff
@@ -620,7 +620,7 @@ void Server::onReadable(void *vp, int status, int events)
     memcpy(src, socketData->spill, socketData->spillLength);
     FD fd;
     uv_fileno((uv_handle_t *) p, (uv_os_fd_t *) &fd);
-    int length = socketData->spillLength + read(fd, src + socketData->spillLength, BUFFER_SIZE - socketData->spillLength);
+    int length = socketData->spillLength + recv(fd, src + socketData->spillLength, BUFFER_SIZE - socketData->spillLength, 0);
```
alexhultman: Good stuff.

alexhultman:
C++/CLI is the .NET monstrosity right? Why is there no standard mutex and why should we care to work around Microsoft flaws?

peters:
> C++/CLI is the .NET monstrosity right?

Yup.

> Why is there no standard mutex

The CLR cannot guarantee that a managed thread runs on a dedicated operating system thread. Technically you could end up two threads sharing the same underlying OS thread and that would really fuck things up :dash: 

> why should we care to work around Microsoft flaws?

That is up to you bro, but in this case it should not make any difference since we are using libuv's mutex implementation so you could choose to accept it based on the assumption of improved portability :laughing: 

But anyway, I'll keep sending you PR's and you can just cherry-pick whatever you want :+1:

src/uWS.cpp
peters: Hmm, looks like a brainfart on my end then. I thought it was a simple POSIX mapping issue :/

alexhultman:
I'm sorry for being a dick about it but I have quite the distaste for non-standard & proprietary Microsoft-only extensions. If their software is broken they need to fix it. I don't want more dependency on libuv than what is minimally needed. I want to easily be able to swap to mTCP whenever I feel that is needed. Then I need to toss out libuv.

peters:
> I'm sorry for being a dick about it but I have quite the distaste for non-standard & proprietary Microsoft-only extensions. 

That's totally understandable, you're the maintainer :) On my part it sucks even more because I need to write the wrapper in c++/cli :laughing: in order to use it in a managed library (C#).

> I don't want more dependency on libuv than what is minimally needed.

Would you accept a Windows mutex implementation then? It's about 50 LOH and if you are compiling for another platform it can fallback to `std::mutex`.

> I want to easily be able to swap to mTCP whenever I feel that is needed. Then I need to toss out libuv.

Interesting library, but seem like a lot of work to integrate into current implementation :)

alexhultman:
But the project does compile to a shared library, aka DLL. You should be able to just use the DLL straight up from C# without any wrapper.

alexhultman:
Ah, right, the native code cannot callback the managed code. Yeah well then you have quite the project.

peters:
I've updated the mutex implementation, care to take a look ? :)

peters:
> Ah, right, the native code cannot callback the managed code. Yeah well then you have quite the project.

Haha, oh yeah. It's actually not that difficult since you write the unit tests in C# :)

src/uWS.cpp
```diff
@@ -18,10 +18,16 @@ using namespace std;
...
+#include <cassert>
+inline FD dup(FD socket)
+{
+ WSAPROTOCOL_INFO pi;
+ auto rv = WSADuplicateSocket(socket, ::GetCurrentProcessId(), &pi); 
+ assert(rv == 0);
+ auto socketDuped = WSASocket(AF_INET, SOCK_STREAM, IPPROTO_TCP, &pi, 0, WSA_FLAG_OVERLAPPED); 
+ assert(socketDuped > 0);
+ return socketDuped;  
+}
```
alexhultman: This is a very nice addition, I'll take a look at this tomorrow. I'm super tired today.

peters: :+1: :beers:

alexhultman:
But the problem is not really solved with that mutex. From looking around the problem seems to be that you cannot use any Win32 threading functions in C++/CLI. It complains for std::mutex, sure, but it is still broken even if it doesn't complain about it. Libuv mutex makes use of Win32 critical sections just like this code you wrote. A Win32 critical section will get the OS thread id, just like std::mutex. Any C++/CLI code should use the .NET threading functions instead, according to some brief StackOverflow Googling.

alexhultman:
Since Microsoft supply the mutex implementation themselves, why would they be so stupid to not just ifdef inside of it and use proper .NET threading when compiled as C++/CLI? It's their header file, their problem. Why are they moving their own problems out of their software and into all dependants? They should have just fired the whole team working on that header.

peters:
Hehe, the thing is that this is just a trick in order to use the header file in the c++ cli project. uWS is statically linked into the final assembly. But I guess people could missunderstand by the name of the define, maybe better to call it WIN32_THREADS?

peters:
That's why there is a #pragma managed(push, off)

peters:
@alexhultman I've updated the PR based on the latest feedback. Is this something you can live with? :)

src/uWS.h
```diff
@@ -1,11 +1,30 @@
...
+#if HAVE_WIN32_CPP_CLI
+#define USE_WIN32_THREADS 1
+#else
+#define USE_WIN32_THREADS 0
+#endif
+#endif
+
 #include <cstddef>
 #include <functional>
-#include <string>
```
peters: @alexhultman I removed an unused include.

src/uWS.cpp
```diff
@@ -1,3 +1,7 @@
+#if HAVE_WIN32_CPP_CLI
```
peters: @alexhultman Added a header guard preventing people from fucking up by embedding uWS directly into their project. Though i seriously doubt there will be anyone but me :laughing:

alexhultman:
I will take a look at this when I have fixed the problem with the primus tests.

alexhultman:
I did commit the recv fix myself, that was too simple to just write in and commit. If you want, could you make a pull request with only the dup? And maybe fix the style to match?

peters:
I'm on it.

PR #48, opened 2016-04-12
Add support for c++/cli

peters:
I cleaned up the revision history and created a new branch for this feature.

History of previous PR is available here: https://github.com/alexhultman/uWebSockets/pull/42

/cc @alexhultman

### DIFF
--- a/src/uWS.cpp
+++ b/src/uWS.cpp
@@ -1,3 +1,7 @@
+#if HAVE_WIN32_CPP_CLI
+#error "c++/cli does not support WIN32 threading model"
+#endif
+
 #include "uWS.h"
 using namespace uWS;
 
@@ -40,8 +44,6 @@ struct WindowsInit {
 #include <openssl/bio.h>
 #include <openssl/evp.h>
 #include <openssl/buffer.h>
-
-#include <uv.h>
 
 enum SendFlags {
     SND_CONTINUATION = 1,
--- a/src/uWS.h
+++ b/src/uWS.h
@@ -1,11 +1,30 @@
 #ifndef UWS_H
 #define UWS_H
 
+#if defined(_WIN32) && defined(__cplusplus_cli)
+#define HAVE_WIN32_CPP_CLI 1
+#pragma managed(push, off)
+#else
+#ifndef HAVE_WIN32_CPP_CLI 
+#define HAVE_WIN32_CPP_CLI 0
+#endif
+#endif
+
+#ifndef USE_WIN32_THREADS
+#if HAVE_WIN32_CPP_CLI
+#define USE_WIN32_THREADS 1
+#else
+#define USE_WIN32_THREADS 0
+#endif
+#endif
+
 #include <cstddef>
 #include <functional>
-#include <string>
 #include <queue>
+
+#if !USE_WIN32_THREADS
 #include <mutex>
+#endif
 
 #ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN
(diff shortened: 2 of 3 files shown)


src/uWS.h
```diff
@@ -1,11 +1,30 @@
...
+#if HAVE_WIN32_CPP_CLI
+#define USE_WIN32_THREADS 1
+#else
+#define USE_WIN32_THREADS 0
+#endif
+#endif
+
 #include <cstddef>
 #include <functional>
-#include <string>
```
peters: @alexhultman Unused include.

src/uWS.h
```diff
@@ -1,11 +1,30 @@
 #ifndef UWS_H
 #define UWS_H
 
+#if defined(_WIN32) && defined(__cplusplus_cli)
+#define HAVE_WIN32_CPP_CLI 1
```
alexhultman: Why aren't you just using #ifdef __cplusplus_cli directly?

src/uWS.cpp
```diff
@@ -1,3 +1,7 @@
+#if HAVE_WIN32_CPP_CLI
+#error "c++/cli does not support WIN32 threading model"
+#endif
```
alexhultman: So you are checking if the .cpp is being compiled with /clr here?

src/uWS.h
peters: No reason in particular, just to imply that this was Windows related. Shall i remove it?

src/uWS.cpp
peters: Correct.

src/uWS.h
```diff
@@ -1,11 +1,30 @@
...
+#else
+#ifndef HAVE_WIN32_CPP_CLI 
+#define HAVE_WIN32_CPP_CLI 0
+#endif
+#endif
+
+#ifndef USE_WIN32_THREADS
+#if HAVE_WIN32_CPP_CLI
+#define USE_WIN32_THREADS 1
+#else
```
alexhultman: USE_WIN32_THREADS is the same as __cplusplus_cli

src/uWS.h
```diff
@@ -14,8 +33,51 @@
...
+#endif
+  
+  MutexImpl(const MutexImpl& other) = delete;
+  MutexImpl& operator=(const MutexImpl& other) = delete;
+
+public:
+  MutexImpl() {
+#if USE_WIN32_THREADS
+    InitializeCriticalSection(&m_mutex);
+#endif
```
alexhultman: Like I said, from Googling the problem, StackOverflow says that you should use .NET threading when in C++/CLI. This is using the Win32 threading functions.

src/uWS.h
peters: Some people use ancient visual studio versions and in order to use `mutex` you need at least VS 2012. If you don't care about those (I myself certainly don't) then I'll gladly remove it :)

src/uWS.h
```diff
@@ -1,11 +1,30 @@
 #ifndef UWS_H
 #define UWS_H
 
+#if defined(_WIN32) && defined(__cplusplus_cli)
+#define HAVE_WIN32_CPP_CLI 1
+#pragma managed(push, off)
```
alexhultman: Is this marking the current file as unmanaged?

peters: Correct.

src/uWS.h
peters: I'll fix it.

alexhultman:
From this petition: https://visualstudio.uservoice.com/forums/121579-visual-studio-2015/suggestions/3814789-support-c-11-concurrency-header-files-in-c-cli

I can navigate to comments from Microsoft about how they have "unblocked" some C++11 STL features in Visual Studio 2015. What this means is really that Microsoft has now finally *implemented* parts of C++11 for their C++/CLI platform. In other words, this is up to Microsoft to fix and this is not just about mutex but also thread, atomic, condition_variable, etc. I cannot accept workarounds based on Microsofts own incompetence with the C++ standard. This code is not set in stone and will change over time - I cannot have workarounds for standard parts of the C++ language based on Microsofts incapacity to follow ISO standards. Microsoft has always been at least 7 years behind everyone else when it comes to C++ and I cannot have them stall my use of standard features. When I start using C++17 it will break in other places and then we have to fix that up.

peters:
That's okay! I'll just maintain a separate branch and try to stay on top of your changes. Thank you for taking the time to review my changes at least :)

alexhultman:
Thanks for the interest

PR #50, opened 2016-04-13
Ensure all properties exist during construction time.

ronkorving:
Not doing so is a V8 performance anti-pattern.

See [Hidden Classes](http://www.html5rocks.com/en/tutorials/speed/v8/)

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -6,6 +6,8 @@ const uws = require(`./uws_${process.platform}_${process.versions.modules}`);
 const NativeServer = uws.Server;
 const EE_ERROR = "registering more than one listener to websocket";
 
+function noop() {}
+
 class Socket {
     /**
      * Creates a Socket instance.
@@ -16,6 +18,8 @@ class Socket {
     constructor(nativeSocket, server) {
         this.nativeSocket = nativeSocket;
         this.server = server;
+        this.messageCallback = noop;
+        this.disconnectionCallback = noop;
     }
 
     /**
@@ -26,16 +30,17 @@ class Socket {
      */
     on(eventName, f) {
         if (eventName === 'message') {
-            if (this.messageCallback) {
+            if (this.messageCallback !== noop) {
                 throw Error(EE_ERROR);
             }
             this.messageCallback = f;
         } else if (eventName === 'close') {
-            if (this.disconnectionCallback) {
+            if (this.disconnectionCallback !== noop) {
                 throw Error(EE_ERROR);
             }
             this.disconnectionCallback = f;
         }
+        return this;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


alexhultman:
The class type of the object will transition when setting the callbacks though..

ronkorving:
Yeah, I was thinking it could be best to not set it to undefined, but to `noop` given `function noop() {}`. We can simply compare against that noop reference instead of undefined when we need to.

alexhultman:
I will probably test to write the whole ws wrapper in C++ and compare the memory usage, that way you get 100% control over V8 and how things are stored.

ronkorving:
I've gone and updated this commit anyway.

nodejs/dist/uws.js
```diff
@@ -110,15 +119,13 @@ class Server extends EventEmitter {
...
-                    socket.disconnectionCallback();
-                }
+                socket.disconnectionCallback();
                 /* make sure to clear any set data */
                 this.nativeServer.setData(nativeSocket);
             });
 
             this.nativeServer.onMessage((nativeSocket, message, binary, socket) => {
-                if (socket.messageCallback) {
+                if (socket.messageCallback !== noop) {
```
ronkorving: Technically, we don't need this if-statement, since the callbacks are always callable, but just in case to avoid calling `message.toString()` if there is no real listener, I kept this if-statement. We could of course say "we always expect there to be relevance in listening for messages, so the overhead of the if-statement is more painful than the overhead of `message.toString()` being called for nothing.

alexhultman:
If you have a noop set then you don't need to check before calling it. If you remove the check before calling the callbacks then I would like it :)

nodejs/dist/uws.js
alexhultman: Exactly.

nodejs/dist/uws.js
```diff
@@ -25,17 +29,22 @@ class Socket {
      * @param {Function} f
      */
     on(eventName, f) {
+        if (typeof f !== 'function') {
+            throw new TypeError('Event listener must be a function');
+        }
```
alexhultman: I don't like this check though. JavaScript is 100% type-unsafe and if they call us with a space ship that's their problem.

ronkorving:
I added a comment about that exactly as you were writing that :)

nodejs/dist/uws.js
ronkorving: LOL. I like your choice of words. Alright, it's your baby, I'll remove it. But personally I'm a big fan of making things more debuggable especially when this is not in a hot code path.

alexhultman: I know, its probably not going to make any difference but it's just the whole thing with JavaScript making me so angry. The language is just pure crap and broken by design, making the programmer ignore types, but then also having to write checks that the compiler should normally have done had it been a sane language. #flame

alexhultman:
I like this, but like you say the `message.toString()` is questionable. I've looked a lot at just passing a String when the type is text but V8 is not really that good with strings (you cannot build a v8::String from utf-8, only from utf-16 with zero-copy.

alexhultman:
I guess it is very strange to have an open socket with no attached message listener, right?

ronkorving:
Cheers :+1:

PR #51, opened 2016-04-13
Include only required files in npm package

lpinca:
The current npm package contains a png file with the logo. It is very small but still unnecessary.
This patch ensures that only required files will be included in the npm package.


alexhultman:
The image is needed for the README.md on NPM

lpinca:
It is pulled from GitHub if I'm not wrong.

ronkorving:
It's not. It just says "logo.png".

lpinca:
https://www.npmjs.com/package/omega-supreme, logo is not included in the package.

lpinca:
Instead of using a relative link in the README we can use an absolute URL like "https://raw.githubusercontent.com/alexhultman/uWebSockets/master/logo.png".

Thoughts?

alexhultman:
I don't think this is an issue. Having a 9 kb logo supplied with the install (which is about 200 kb) could be seen as bloat, but it could also be seen as a brand. The logo sticks out and makes a statement. Plus, NPM servers would not have to rely on GitHub (this is ofc purely theoretical as the traffic is negligible). One could argue that the LICENSE file is unnecessary and could be fetched from GitHub also.

lpinca:
Fine, no problem.

PR #55, opened 2016-04-14
Pass an error instance to the error listener

lpinca:
Right now the error listener is invoked with a string.
It is usually expected to receive an error instance in the error listener.


alexhultman:
Nice nice :)

PR #70, opened 2016-04-24
Add an optional callback to socket.send

LordMajestros:
Add an optional callback to socket.send.
This is for ws compatibility.

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -99,7 +99,7 @@ class Socket {
      */
     send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if(typeof options === 'function') {
+        if (typeof options === 'function') {
             cb = options;
             options = {};
         }
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -101,7 +101,7 @@ class Socket {
         /* ignore sends on closed sockets */
         if (typeof options === 'function') {
             cb = options;
-            options = {};
+            options = null;
         }
         if (!this.nativeSocket) {
             return cb && cb(new Error('not opened'));
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -94,14 +94,23 @@ class Socket {
      *
      * @param {String|Buffer} message The message to send
      * @param {Object} options Send options
+     * @param {Function} cb optional callback
      * @public
      */
-    send(message, options) {
+    send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if (!this.nativeSocket) return;
+        if (!this.nativeSocket){ 
+            if(cb){
+                cb('Socket closed');
+            }
+            return;
+        }
 
         const binary = options && options.binary || false;
         this.server.nativeServer.send(this.nativeSocket, message, binary);
+        if(cb){
+            cb(null);
+        }
     }
 
     /**
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -99,6 +99,10 @@ class Socket {
      */
     send(message, options, cb) {
         /* ignore sends on closed sockets */
+        if(typeof options==='function'){
+            cb = options;
+            options = {};
+        }
         if (!this.nativeSocket){ 
             if(cb){
                 cb('Socket closed');
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -104,17 +104,12 @@ class Socket {
             options = {};
         }
         if (!this.nativeSocket) {
-            if(cb) {
-                cb('Socket closed');
-            }
-            return;
+            return cb && cb(new Error('not opened'));
         }
 
         const binary = options && options.binary || false;
         this.server.nativeServer.send(this.nativeSocket, message, binary);
-        if(cb) {
-            cb(null);
-        }
+        return cb && cb(null);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


nodejs/dist/uws.js
```diff
@@ -94,14 +94,27 @@ class Socket {
...
+     * @param {Function} cb optional callback
      * @public
      */
-    send(message, options) {
+    send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if (!this.nativeSocket) return;
+        if(typeof options==='function'){
+            cb = options;
+            options = {};
```
alexhultman: This line doesn't do anything?

nodejs/dist/uws.js
```diff
@@ -94,14 +94,27 @@ class Socket {
...
      * @param {String|Buffer} message The message to send
      * @param {Object} options Send options
+     * @param {Function} cb optional callback
      * @public
      */
-    send(message, options) {
+    send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if (!this.nativeSocket) return;
+        if(typeof options==='function'){
```
alexhultman: Formatting, spaces

nodejs/dist/uws.js
```diff
@@ -94,14 +94,27 @@ class Socket {
...
+            cb = options;
+            options = {};
+        }
+        if (!this.nativeSocket){
```
alexhultman: Inconsistent formatting, trailing space

nodejs/dist/uws.js
```diff
@@ -94,14 +94,27 @@ class Socket {
...
+    send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if (!this.nativeSocket) return;
+        if(typeof options==='function'){
+            cb = options;
+            options = {};
+        }
+        if (!this.nativeSocket){ 
+            if(cb){
+                cb('Socket closed');
```
alexhultman: Why is there a string?

alexhultman: Ah, its an error?

JrSchild: Also note that this can be reduced to: `return cb && cb('Socket closed')`.

alexhultman: This seems to be represented by `new Error('not opened')` in ws.

nodejs/dist/uws.js
```diff
@@ -94,14 +94,27 @@ class Socket {
...
+            }
+            return;
+        }
 
         const binary = options && options.binary || false;
         this.server.nativeServer.send(this.nativeSocket, message, binary);
+        if(cb){
+            cb(null);
```
alexhultman: Calling the cb after the async function is probably not 100% correct but the documentation is really ambiguous and there is really no way to tell when the message has really been "sent".

nodejs/dist/uws.js
LordMajestros: Yes it is, will change

nodejs/dist/uws.js
```diff
@@ -99,20 +99,20 @@ class Socket {
      */
     send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if(typeof options==='function'){
+        if(typeof options === 'function') {
```
alexhultman: this if

nodejs/dist/uws.js
```diff
@@ -99,20 +99,20 @@ class Socket {
...
             cb = options;
             options = {};
         }
-        if (!this.nativeSocket){ 
-            if(cb){
+        if (!this.nativeSocket) {
```
alexhultman: has different formatting than this one

nodejs/dist/uws.js
LordMajestros: True but I'm uncertain of where to put it. Unless we can provide a callback to the native send function that calls cb when it's sent.

nodejs/dist/uws.js
```diff
@@ -99,20 +99,20 @@ class Socket {
      */
     send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if(typeof options==='function'){
+        if(typeof options === 'function') {
             cb = options;
             options = {};
```
alexhultman: I'm pretty sure this line does nothing

nodejs/dist/uws.js
alexhultman: It should be fine as the docs doesn't specify, also, it's kind of bullshit callback anyways because there is no way to know when things has been sent (especially in libuv where they queue the data in their own buffers and call that a "send").

nodejs/dist/uws.js
LordMajestros: What if options is omitted and the callback passed as the second arg?

LordMajestros: What if options is omitted and the callback passed as the second arg?

nodejs/dist/uws.js
```diff
@@ -94,14 +94,22 @@ class Socket {
...
+     * @param {Function} cb optional callback
      * @public
      */
-    send(message, options) {
+    send(message, options, cb) {
         /* ignore sends on closed sockets */
-        if (!this.nativeSocket) return;
+        if (typeof options === 'function') {
+            cb = options;
+            options = {};
```
alexhultman: Maybe we could make this `options = null` instead. I'm pretty sure {} creates a new object which is not neccesary. We only want `options` to evaluate to `false`.

LordMajestros: True, done.

alexhultman:
:+1:

PR #84, opened 2016-05-02
Use $(CXX) and add build instructions for benchmarks

### LINKED ISSUE #80: benchmarks building is limited to gcc and its instructions are not well documented and failing.
## Compiler limitation

benchmarks building is limited to gcc , but I'm using clang(and `CXX` is set to `clang++`), so I needed to modify the Makefile in `benchmarks` to something like:

```make
CXX ?= g++
default:
	$(CXX) -std=c++11 -O3 bench1.cpp -s -o bench1 -lpthread 
# the rest is similar
```

## Dependency issues

bench1 builds fine without any dependencies.

bench2 requires [WebSocket++](https://github.com/zaphoyd/websocketpp) which in turn requires boost >> 1.41 (my production server default version of boost, I don't know the lower limit of version, but I just switched to the latest version)

bench3 and wsPP are similar to bench2.

uWS refers to missing `../uWS.cpp` and `../utf8.cpp`, the former is possibly just `./uWS.cpp`, but the latter is not present.

utensil:
Fixes #80 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+CMakeCache.txt
+CMakeFiles/
+Makefile
+cmake_install.cmake
+examples/CMakeFiles/
+examples/Makefile
+examples/cmake_install.cmake
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,4 +1,19 @@
-These are the benchmarks used in testing the different WebSocket server implementations. I use WebSocket++ as clients, since that is the fastest client library that I know of, and since you can run multiple instances of the client, it shouldn't really matter what client you use as long as the server is being stressed fully.
+# uWebSockets Benchmarks
+
+## Building Instructions
+
+First, install dependencies:
+
+* `bench1` depends only uWebSockets.
+* `bench2`, `bench3` and `wsPP` also depends on:
+  - [WebSocket++](https://github.com/zaphoyd/websocketpp) which is header-only, you can run `make install` in its directory to install the headers to system include path
+  - [boost](http://www.boost.org/) which must be built and installed, you could follow the [instructions for Linux/Mac](http://www.boost.org/doc/libs/1_60_0/more/getting_started/unix-variants.html) or [instructions for Windows](http://www.boost.org/doc/libs/1_60_0/more/getting_started/windows.html)
+
+Then just run `make`. You can change your environment vairable `CXX` to specify a C++11 compliant compiler, or it would default to `g++`. 
+
+## Benchmark Tips
+
+These are the benchmarks used in testing the different WebSocket server implementations. I use [WebSocket++](https://github.com/zaphoyd/websocketpp) as clients, since that is the fastest client library that I know of, and since you can run multiple instances of the client, it shouldn't really matter what client you use as long as the server is being stressed fully.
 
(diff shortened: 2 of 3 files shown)


benchmarks/README.md
```diff
@@ -1,4 +1,19 @@
-These are the benchmarks used in testing the different WebSocket server implementations. I use WebSocket++ as clients, since that is the fastest client library that I know of, and since you can run multiple instances of the client, it shouldn't really matter what client you use as long as the server is being stressed fully.
+# uWebSockets Benchmarks
+
+## Building Instructions
+
+First, install dependencies:
+
+* `bench1` depends only uWebSockets.
```
alexhultman: typo, also - there is not really a dependency on µWS.

benchmarks/README.md
```diff
@@ -1,4 +1,19 @@
...
+* `bench2`, `bench3` and `wsPP` also depends on:
+  - [WebSocket++](https://github.com/zaphoyd/websocketpp) which is header-only, you can run `make install` in its directory to install the headers to system include path
+  - [boost](http://www.boost.org/) which must be built and installed, you could follow the [instructions for Linux/Mac](http://www.boost.org/doc/libs/1_60_0/more/getting_started/unix-variants.html) or [instructions for Windows](http://www.boost.org/doc/libs/1_60_0/more/getting_started/windows.html)
```
alexhultman: I don't really agree with this - for me it was just a matter of sudo dnf install boost-devel. I also know that the same applies for OS X where you use homebrew instead

benchmarks/README.md
```diff
@@ -1,4 +1,19 @@
-These are the benchmarks used in testing the different WebSocket server implementations. I use WebSocket++ as clients, since that is the fastest client library that I know of, and since you can run multiple instances of the client, it shouldn't really matter what client you use as long as the server is being stressed fully.
+# uWebSockets Benchmarks
```
alexhultman: I would make the caption "Benchmarks" as these are applicable to any WS implementation.

.gitignore
```diff
@@ -0,0 +1,7 @@
+CMakeCache.txt
+CMakeFiles/
+Makefile
+cmake_install.cmake
+examples/CMakeFiles/
+examples/Makefile
+examples/cmake_install.cmake
```
alexhultman: I don't want a .gitignore

lpinca: This is useful, I also wanted to add it, when playing with the repo it's kinda boring to deal with untracked stuff.

benchmarks/README.md
utensil: Oh, sorry for the typo and misunderstanding. I saw your fix at efb98947eb65cb61d0e3363d0ae9e03f54254729 .

benchmarks/README.md
utensil: Yes. For most people with a reasonably modern distribution would enjoy the convenience that the package managers provides. I was in the context of my old distribution on production server and neglected to mention.

.gitignore
utensil: I don't even know how to commit without a `.gitignore` as there are many untracked CMake generated files.

Also it's better to further add more entries into `.ignore` just like [the standard template that Github provides for node.js projects](https://github.com/github/gitignore/blob/master/Node.gitignore).

PR #85, opened 2016-05-04
Preliminary work on binding.gyp

kapouer:
This file won't work "as is":
- source files must be copied in dist dir
- uws.js does not load `uws.node` but `uws_xx_yy.node`
but the bulk is there.

### DIFF
--- a/nodejs/dist/binding.gyp
+++ b/nodejs/dist/binding.gyp
@@ -1,0 +1,30 @@
+{
+  'targets': [
+    {
+      'target_name': 'uws',
+      'sources': [
+        'src/uWS.cpp',
+        'src/addon.cpp'
+      ],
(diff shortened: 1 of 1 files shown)


alexhultman:
Thanks, but I'm still uncertain about this. There is really nothing hindering one from just getting the sources from github and compiling them. My initial thought was I wanted to provide prebuilt binaries over NPM that just works. NPM as a transport rather than a complete infrastructure.

kapouer:
Hey then why not distribute two npm packages ? `uws` and `uws-source` or something like that ?
I can fix the binding.gyp so that it names the `uws.node` binary according to the initial naming you're using, this way `uws.js` won't need a change.

alexhultman:
Yep we could have two packages that could be one solution. That way you coulddetermine what solution people prefer.

PR #87, opened 2016-05-04
Fix possible race condition in handleUpgrade

### LINKED ISSUE #86: Possible race condition in handleUpgrade
Every upgrade request registers a new [`onConnection`](https://github.com/alexhultman/uWebSockets/blob/5793d5813da116b58f28911a2eecf73f34e43495/nodejs/dist/uws.js#L268-L278) handler, but [`server.upgrade`](https://github.com/alexhultman/uWebSockets/blob/5793d5813da116b58f28911a2eecf73f34e43495/nodejs/dist/uws.js#L286) is called asynchronously when the socket is closed.

If there are concurrent requests I think that it is possible that the wrong `onConnection` handler will be used.

For example if there are 2 concurrent requests they will both use the `onConnection` handler of the second one.

I haven't tested this so I may be totally wrong.

lpinca:
This sets the `onConnection` listener only when the socket has been closed to prevent possible race conditions.

Closes #86.

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -264,28 +264,29 @@ class Server extends EventEmitter {
      * @public
      */
     handleUpgrade(request, socket, upgradeHead, callback) {
-        /* register a special connection handler */
-        this.nativeServer.onConnection((nativeSocket) => {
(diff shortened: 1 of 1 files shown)


alexhultman:
Thanks

Issue #75, opened 2016-04-26
Amazon/RedHat issue - Error: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.14' not found


jondubois:
While testing on Amazon (based on Redhat), this error came up:

Error: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.14' not found (required by /home/ec2-user/sc-test/node_modules/uws/uws_linux_47.node)
    at Error (native)
    at Object.Module._extensions..node (module.js:422:18)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/ec2-user/sc-test/node_modules/uws/uws.js:5:13)
    at Module._compile (module.js:397:26)
    at Object.Module._extensions..js (module.js:404:10)
    at Module.load (module.js:343:32)

Looks like the c compiler of Amazon's Distro (and maybe Redhat too) is not supported?

alexhultman:
Amazon uses ancient compiler versions. Either I have to bundle my version of libstdc++ or link statically.

jondubois:
Yeah, I guess linking statically or bundling might be worthwhile... If it doesn't add too much filesize. Portability is pretty important when it comes to Node.js.

alexhultman:
Another solution could possibly be to compile it using an older compiler, I think that will solve itself if I opt for using Travis because they are also using ancient versions (their Ubuntu version is from the stone age).

alexhultman:
#54

alexhultman:
I don't think it's a good idea bundling libstdc++. It's 1.6 MB large and should really not be bundled IMO. I still think the best choice is to just use an older C++ compiler. If you look at Node.js itself, it also depends on libstdc++ but they don't ship it - they simply compiled Node.js using an older compiler.

But I still need to solve this, I use Fedora which comes with very up to date compilers and I'm not going to swap to some crappy ancient Ubuntu shit, so I need to set up Travis or any other service to get this working.

alexhultman:
@jondubois What version of RHEL are you running? Version 7 could be a good target, right?

jondubois:
@alexhultman I'm not sure which version of RHEL my Amazon Linux is based on.
My GCC version is 4.4.6. This image is a few years old.

[...]

PR #88, opened 2016-05-04
Distribute two npm modules: source and binary

kapouer:
## Makefile changes
To develop, just `make`.
To prepare a release, `VERSION=0.3.2 make prepare` and check the content of uws-binary, uws-source folders.
To publish, `VERSION=0.3.2 make publish` will publish both.
Also note the stamp on targets/ to avoid redownloading.

## package.json changes
The `uws` package tells which nodejs versions and os it exactly supports.
The `uws-source` package is a little less strict by essence.
I alos finished the work on binding.gyp so flags are set for darwin, and filename is the one expected by uws.js. I think it's a bad idea to load a versioned filename, because the error message will be misleading (missing module). Node already has a module version mismatch check and appropriate warning - but i didn't change that.


nodejs/uws-source/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "uws-source",
+  "version": "0.3.1",
+  "scripts": {
+    "build": ""
+  },
+  "main": "uws.js",
+  "description": "Highly scalable WebSocket server library - node-gyp compilable version",
+  "engines": {
+    "node": ">=4.4 <=6.0"
```
alexhultman: The source will compile from 4.x to 6.x

nodejs/uws-binary/package.json
```diff
@@ -5,7 +5,11 @@
     "build": ""
   },
   "main": "uws.js",
-  "description": "Highly scalable WebSocket server library",
+  "description": "Highly scalable WebSocket server library - binaries for Linux/Darwin",
+  "engines": {
+    "node": "4.4.3 5.11.0 6.0.0"
```
alexhultman: Same goes here, it is actually compatible with 4.x, 5.x, 6.x - only the major release numbers matter

alexhultman:
This looks pretty nice. If the uws-source becomes more popular then I can simply publish it under the name of uws - it can serve as an alternative distribution in the mean time.

nodejs/uws-binary/package.json
kapouer: ok, fixing those

kapouer:
I'd like to propose a different solution. I don't like that one, because it might confuse users a little too much.
The idea would be to publish the uws-source package with the binaries precompiled (as if we merged the two sets of files), then at the moment of installation, i can setup a script that tries to run `node-gyp rebuild` but doesn't complain in case it fails for any reason (like missing node-gyp and dev files) - it would fall back to the precompiled files nicely.

nodejs/uws-binary/package.json
```diff
@@ -5,7 +5,11 @@
     "build": ""
   },
   "main": "uws.js",
-  "description": "Highly scalable WebSocket server library",
+  "description": "Highly scalable WebSocket server library - binaries for Linux/Darwin",
```
alexhultman: Maybe this description should be the original one.

nodejs/uws-source/binding.gyp
```diff
@@ -6,7 +6,14 @@
         'src/uWS.cpp',
         'src/addon.cpp'
       ],
-      'cflags_cc': [ '-std=c++11', '-fexceptions' ]
+      'conditions': [
+        ['OS=="linux"', {
+          'cflags_cc': [ '-std=c++11', '-fexceptions' ]
+        }],
+        ['OS=="darwin"', {
+          'cflags_cc': [ '-stdlib=libc++', '-mmacosx-version-min=10.7', '-undefined' ]
```
alexhultman: I think you need exceptions here too.

kapouer: Huu right it's CPP_SHARED *and* CPP_OSX, not *or*.

kapouer:
Here it goes, with a nice message "Falling back to binaries" in case of node-gyp failure.

alexhultman:
This is awesome. If this works we shouldn't even need the uws-source? Just put everything in uws?

nodejs/dist/binding.gyp
```diff
@@ -6,7 +6,12 @@
         'src/uWS.cpp',
         'src/addon.cpp'
       ],
-      'cflags_cc': [ '-std=c++11', '-fexceptions' ]
+      'cflags_cc': [ '-std=c++11', '-fexceptions' ],
```
alexhultman: We need to negate the -fno-exceptions also, add `'cflags_cc!': [ '-fno-exceptions' ]`

nodejs/dist/binding.gyp
```diff
@@ -6,7 +6,12 @@
...
+      'conditions': [
+        ['OS=="darwin"', {
+          'cflags_cc': [ '-stdlib=libc++', '-mmacosx-version-min=10.7', '-undefined' ]
```
alexhultman: '-stdlib=libc++' is the only flag we need here, as we target our own system (I'm also 99% sure node-gyp automatically adds the -undefined dynamic_lookup flag automagically.

nodejs/dist/package.json
```diff
@@ -2,10 +2,14 @@
   "name": "uws",
   "version": "0.3.1",
   "scripts": {
-    "build": ""
+    "preinstall": "node-gyp rebuild && exit 0; echo 'Falling back to precompiled binaries';"
```
alexhultman: Is this really going to work? If I type node-gyp in my terminal it asks me if I should install it and it expects y/N keyboard input to continue.

nodejs/dist/package.json
```diff
@@ -2,10 +2,14 @@
...
   },
   "main": "uws.js",
   "description": "Highly scalable WebSocket server library",
+  "engines": {
+    "node": "4.x 6.x"
```
alexhultman: Is this correct? What about 5.x?

nodejs/dist/package.json
alexhultman: Maybe some kind of check with `which node-gyp` or something like that could work? How complex can this preinstall script be?

alexhultman: Never mind, I think it always finds it when run through NPM.

nodejs/dist/package.json
kapouer: You're right i cannot get ir right, apparently

alexhultman: :)

nodejs/dist/package.json
kapouer: Yes, npm controls the env and actually somewhat parses the command here.
For example i couldn't do `node-gyp rebuild || echo "xxx"`

kapouer:
Updated with your remarks.
I choose to simply put engine node >= 4 because it's simpler to maintain and contains the main information (that it is incompatible with < 4).

kapouer:
A word about documentation: the right way to trigger a build is "npm install".

alexhultman:
I'm sorry but the fallback doesn't work for me. If I uninstall my gcc I get this:

```
[alexhultman@localhost dist]$ npm install

> uws@0.3.2 install /home/alexhultman/uWebSockets/nodejs/dist
> node-gyp rebuild

make: Entering directory '/home/alexhultman/uWebSockets/nodejs/dist/build'
  CXX(target) Release/obj.target/uws/src/uWS.o
make: g++: Command not found
uws.target.mk:89: recipe for target 'Release/obj.target/uws/src/uWS.o' failed
make: *** [Release/obj.target/uws/src/uWS.o] Error 127
make: Leaving directory '/home/alexhultman/uWebSockets/nodejs/dist/build'
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/opt/node-v4.2.6-linux-x64/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:276:23)
gyp ERR! stack     at emitTwo (events.js:100:13)
gyp ERR! stack     at ChildProcess.emit (events.js:185:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:204:12)
gyp ERR! System Linux 4.2.5-201.fc22.x86_64
gyp ERR! command "/opt/node-v4.2.6-linux-x64/bin/node" "/opt/node-v4.2.6-linux-x64/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /home/alexhultman/uWebSockets/nodejs/dist
gyp ERR! node -v v5.11.0
gyp ERR! node-gyp -v v3.3.1
gyp ERR! not ok 

npm WARN uws@0.3.2 license should be a valid SPDX license expression
npm ERR! Linux 4.2.5-201.fc22.x86_64
npm ERR! argv "/opt/node-v4.2.6-linux-x64/bin/node" "/opt/node-v4.2.6-linux-x64/bin/npm" "install"
npm ERR! node v5.11.0
npm ERR! npm  v3.8.6
npm ERR! code ELIFECYCLE
npm ERR! uws@0.3.2 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the uws@0.3.2 install script 'node-gyp rebuild'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the uws package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs uws
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls uws
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /home/alexhultman/uWebSockets/nodejs/dist/npm-debug.log
```

kapouer:
That should be fixable, let me check

kapouer:
That should do it. I'm using an old npm version with a different behavior w.r.t. preinstall/install. Latest npm docs gave the right answer.

kapouer:
Rebased over 0.3.3

kapouer:
Is something holding you ?

alexhultman:
Sorry, I appreciate the work but I'm still considering it. The main issue was #75 which I recently closed with this commit: 86583d1bf570c5b72ce66e9cb892fdfddca56c5c. Now it seems to work out of the box on Linux and OS X on all the supported versions of Node.js.

I tried this updated package.json & binding.gyp and the experience is still not that great when it fails. You still get tons of error messages and hints about broken package, etc. If I were to merge this it wouldn't really solve any real problem, just add an extra step in the install process that might fail (even though it doesn't *really* fail).

I think this change could be used for a separate source-only distribution but I'm not overly attached to its current state. I want the experience to be flawless - it should just work out of the box.

kapouer:
Latest commit address the problem with failing node-gyp during install. Now it's really a few non-worrying line in case it falls back to binaries.

The rest of your concerns i cannot address. What i proposed with that PR is what expect a vast majority of npm users when they do `npm install someaddon`. A smoother experience, while still retaining ability to hack and rebuild from the npm package (as it is the case for most addons in npm), is provided by tools like `node-pre-gyp` but you said you want to avoid depending on more tools.

alexhultman:
I already tried redirecting stderr to /dev/null but it still printed errors. It seems NPM itself is printing these errors. If I were to merge this PR some users would begin to get these errors which is not something I want to happen.

No offense but I don't see the problem here. The module already works and if someone wanted to get the sources they can get them from here. When I develop I simply have the git repo, I don't need NPM at all - it's just a way of distributing the end result.

kapouer:
Here's why i get if i run `npm install` without, say, g++ in the path:
```
npm install

> uws@0.3.3 install /home/dev/Software/github/uWebSockets/nodejs/dist
> node-gyp rebuild 2> /dev/null 1> /dev/null && exit 0; echo 'Falling back to precompiled binaries';

Falling back to precompiled binaries
```

If you want something more silent than that i can even remove "falling back to...".

As for distributing the end result, i think it's a misconception that your builds are "one size fits all", especially when it comes to stdc++ abi compatibility and such. But i'll stop annoying you with that matter and let you realize that later when hordes of angry users come knocking at your door ;)

alexhultman:
This is what I get if I put invalid code in one of the sources:

```
[alexhultman@localhost dist]$ npm install

> uws@0.3.1 install /home/alexhultman/uWebSockets/nodejs/dist
> node-gyp rebuild 2> /dev/null 1> /dev/null && exit 0;


npm ERR! Linux 4.2.5-201.fc22.x86_64
npm ERR! argv "/opt/node-v4.2.6-linux-x64/bin/node" "/opt/node-v4.2.6-linux-x64/bin/npm" "install"
npm ERR! node v5.11.0
npm ERR! npm  v3.8.6
npm ERR! code ELIFECYCLE
npm ERR! uws@0.3.1 install: `node-gyp rebuild 2> /dev/null 1> /dev/null && exit 0;`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the uws@0.3.1 install script 'node-gyp rebuild 2> /dev/null 1> /dev/null && exit 0;'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the uws package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild 2> /dev/null 1> /dev/null && exit 0;
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs uws
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls uws
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /home/alexhultman/uWebSockets/nodejs/dist/npm-debug.log
```

kapouer:
My (stupid) mistake, sorry. Amended.

alexhultman:
That fixes it, how does the exit(0) work? Does it exit npm itself or just the child node-gyp? I mean, will this mess up npm itself if being part of another install or something like that?

alexhultman:
No it works as it should. I tried changing the exit 0 to exit 2 and I got errors from NPM saying the exit status was 2, so it's only the child process. That's super, I will look some more into this and most probably merge. I just want to take a look at the Makefile some more, it contains a few changes that are not really needed IMO.

kapouer:
I'm glad it's ok for you.
I think it'd be better to leave the actual "node-gyp" ouput (maybe only stdout or stderr) instead of shutting it off - or else it's going to be tricky to know what happened in case both compilation and binary fail. Npm would stay silent anyway.

alexhultman:
Yep this actually seems to work better than I thought it could. Thanks for sticking with me. I've heard a lot of negativity towards node-gyp and I have seen bug reports with many comments about how they [some devs] cannot figure out what is wrong from all the NPM errors when all they really need is to install a compiler. The ws devs also made it very clear they didn't enjoy having native dependencies because of the fails.

This binding.gyp + package.json integrates really well with my initial idea of only distributing binaries. I should be able to add some checks to uws.js and throw a nice error message if require('uws') fails due to missing binary. This should also help Windows users since I do not currently pre-compile for Windows. I should also be able to revert the change where I link statically to libstdc++ and libgcc now since the prebuilt binaries are second class citizens now.

alexhultman:
Would you mind opening a new PR with only package.json and binding.gyp? Also, the goal is to support Windows so I think we need to have every compiler flag in its own OS conditional. We can fix the Makefile separately.

kapouer:
I also remove the "os" field in package.json, since you want to target win32 as well;
and let you deal with .npmignore file. Note that .gitignore is enough to prevent `npm publish` from including unwanted files but that works only if the package root is the repo root.

alexhultman:
I always check before I publish and I don't prebuild with node-gyp so the .npmignore shouldn't be needed. I didn't add the .gitignore either, it came in some other PR. I don't like it but it's there.

kapouer:
See https://github.com/alexhultman/uWebSockets/pull/89

kapouer:
n/a

PR #89, opened 2016-05-05
Install npm package using node-gyp and fail quietly

kapouer:


alexhultman:
Looks nice but OSX also needs -std=c++11

alexhultman:
Very nice, thanks

PR #107, opened 2016-05-13
Update binding.gyp

lpinca:
This updates the mac condition in binding.gyp as discussed in https://github.com/alexhultman/uWebSockets/commit/d4e56bd1bef6939781d91121832601d5eaba0b24.
This is the list of generated flags:

```
-O3 -mmacosx-version-min=10.7 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=c++11 -stdlib=libc++
```


alexhultman:
Thanks I will check this later, seems solved now

PR #113, opened 2016-05-16
Spawned Aleph0 monkeys to fix 100% CPU load bug

kapouer:
I don't know why this works. Don't ask me.
In any case, it prevents this test npm module: `hog-uws-webkigtk-uv@0.0.2` to get
a 100% CPU load when closing a connection.

### DIFF
--- a/src/uWS.cpp
+++ b/src/uWS.cpp
@@ -1234,7 +1234,6 @@ void Socket::close(bool force, unsigned short code, char *data, size_t length)
             delete (uv_poll_t *) handle;
         });
 
-        ::close(fd);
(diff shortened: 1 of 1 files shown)


alexhultman:
If this works then you probably have another poll on the same FD, and when I close it this poll starts to trigger all the time due to broken underlying FD. Make sure to check all involved FD's, it is invalid to close an FD before stopping its poll.

alexhultman:
Make sure to track down all usages of the socket being passed to uws, is this socket passed to anything else? Once passing a socket to uws the ownership changes, we own the socket and there should be no other library using it at all.

PR #121, opened 2016-05-20
Fix cleanup on existing http server when closing

131:
Fix proposal for https://github.com/alexhultman/uWebSockets/issues/120

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const url  = require('url');
 const http = require('http');
 //const https = require('https'); // should we ever create an https server?
 const EventEmitter = require('events');
@@ -209,9 +210,16 @@ class Server extends EventEmitter {
                 options.path = '/' + options.path;
             }
 
-            this.httpServer.on('upgrade', (request, socket, head) => {
-                if (!options.path || options.path == request.url.split('?')[0].split('#')[0]) {
-                    if (options.verifyClient) {
+            this._onServerUpgrade = (request, socket, head) => {
+
+                // check for wrong path
+                if (options.path) {
+                  var u = url.parse(request.url);
+                  if (u && u.pathname !== options.path)
+                    return;
+                }
+		
+		                    if (options.verifyClient) {
                         const info = {
                             origin: request.headers.origin,
                             secure: request.connection.authorized !== undefined || request.connection.encrypted !== undefined,
@@ -244,10 +252,10 @@ class Server extends EventEmitter {
                             this.emit('connection', ws);
                         });
                     }
-                } else {
-                    socket.end();
-                }
-            });
+            };
+
+            this.httpServer.on('upgrade', this._onServerUpgrade);
+
         }
 
         this.nativeServer.onDisconnection((nativeSocket, code, message, socket) => {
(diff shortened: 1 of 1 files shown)


131:
See https://github.com/alexhultman/uWebSockets/issues/120

alexhultman:
Thanks, I will look into this later. You have some tabs mixed with the spaces but we can fix that. I'm just really exhausted mentally right now.

nodejs/dist/uws.js
```diff
@@ -209,9 +210,16 @@ class Server extends EventEmitter {
...
-            this.httpServer.on('upgrade', (request, socket, head) => {
-                if (!options.path || options.path == request.url.split('?')[0].split('#')[0]) {
-                    if (options.verifyClient) {
+            this._onServerUpgrade = (request, socket, head) => {
+
+                // check for wrong path
+                if (options.path) {
+                  var u = url.parse(request.url);
+                  if (u && u.pathname !== options.path)
+                    return;
```
lpinca: Shouldn't the socket be closed here?

131: If you have multiple websocket server listening on different paths (options.path) of the same server instance, you must not close the client socket for a path you do not understand/are not registered for.

131: shamelessly inspired by https://github.com/websockets/ws/blob/master/lib/WebSocketServer.js#L164

alexhultman: This exact issue has been raised at Node.js GitHub and they claim it is not supported to listen on different paths as it "makes no sense" to quote themselves. You'll have to take this up with them.

131: I really do not understand where node.js can stand on how to use "upgrade" request. Can you point me somewhere ? 


```
globalapp = MyAppControler.connect(new Websocket());

var liveStreamCam01 = new Websocket("/cam/hallway");//binary data only
liveStreamCam01.pipe(h264decoder).display($("#canvas1"))

var liveStreamCam02 = new Websocket("/cam/parking");//binary data only
liveStreamCam01.pipe(h264decoder).display($("#canvas2"))

globalapp.on('message', function thatHaveNothingToDoWithCameras)
```

alexhultman: https://github.com/nodejs/node/issues/6339

131: Hahaha :smile: no idea such flame was firing here 

As it's not settled (and my guess is that it will not be soon) feel free to align with ws behavior (or not, it's your call !) 

I'll stick in my fork until them :cake:. (make sure to cleanup the handle, though)

alexhultman: I will take a look at this, but not today.

alexhultman: Francois, I agree with you. My quote was meant as sarcasm. I will most probably go with your design, but today I'm juat taking it slow.

nodejs/dist/uws.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const url  = require('url');
```
alexhultman: I like the use of this thing instead of the ugly .split

alexhultman:
I'm closing this PR, will track this issue in the issue report.

PR #131, opened 2016-05-29
fixes problem with invoking a constructor

capaj:


alexhultman:
Thanks, will not force push again for a while :wink:

PR #139, opened 2016-06-03
specify os compatibilty in package.json

silverwind:
This should stop npm from installing a broken module on Windows. The `os` could also be changed to `"os" : [ "!win32" ]` if Solaris and BSD are known to work.


alexhultman:
Hehe, I would argue its not broken - Node.js is :wink:

SEAPUNK:
You mean Windows? ;)

PR #144, opened 2016-06-08
Add client library

jimmyt857:
I haven't added in all of the error cases yet, but here is the basic idea for the client. I figured since it will need to change anyway before being integrated, it wouldn't be an efficient use of time for me to work in every detail yet. Take a look and let me know what you think.

There is a new example (client_echo.cpp) that you can run to see it work trivially.


jimmyt857:
Two things I forgot to mention: the base64 string conversion function you had only works for fixed-size strings, so I added an additional small file that make it more generic (and is needed for the client which needs to convert a different string).

Second, I encountered a bug which seemed to come from your already present code. The internalFragment() function takes a pointer to the message buffer, and that point seems to need to be the beginning of the message payload (after the message header). However, consumeCompleteMessage() and consumeIncompleteMessage() call internalFragment with the pointer to the beginning of the whole message, including the header. I'm wondering if I missed something that makes up for this difference, because I couldn't get my client to work until I changed that.

jimmyt857:
To get the discussion started on how to actually merge this into the library: most of the differences could be handled by adding a boolean isServer to the SocketData object, and then using that to condionally act on the small differences between client and server. The main difference is whether or not a mask is used, and by consequence the size of the header.

alexhultman:
It looks very similar indeed. I don't want to use any runtime branching for this, so instead I would use inline template functions and template structs. You could probably wrap the whole parser into a template function taking the LONG_MESSAGE_HEADER and company as static constant arguments.

alexhultman:
You don't need the base64 thing - just generate the random string directly as base64.

src/uWS_Client.cpp
```diff
@@ -0,0 +1,1371 @@
...
+    unsigned char state = READ_HEAD;
+    unsigned char sendState = FRAGMENT_START;
+    unsigned char fin = true;
+    char opStack = -1;
+    char spill[16]; // can be 14 in size
+    unsigned char spillLength = 0;
+    OpCode opCode[2];
+    unsigned int remainingBytes = 0;
+    char mask[4];
+    Client *client;
```
alexhultman: You don't need SocketData -> Client relation, Client knows its SocketData. You do need some kind of shared buffer holder though (receiveBuffer, sendBuffer, inflateBuffer should not be per client but shared among all clients just like the Server does)

src/uWS_Client.cpp
```diff
@@ -0,0 +1,1371 @@
...
+            Client::closeHandler((Client *) a->data);
+        });
+    }
+}
+
+Client::~Client()
+{
+    delete [] (uint32_t *) receiveBuffer;
+    delete [] sendBuffer;
+    delete [] inflateBuffer;
```
alexhultman: These needs to be part of a shared buffer holder

alexhultman:
Is it okay to work on this on a separate branch for a while? I think it should be pretty simple to merge but it needs to be done manually anyways.

jimmyt857:
Yeah, a separate branch is fine. Adding templates to the functions is going to affect pretty much every single function, and many of them will need to be moved into the header file because of this. Are you ok with that?

src/uWS_Client.cpp
jimmyt857: The SocketData-> Client is copied over from the SocketData -> Server relation in the existing code. It's needed in several places, for example consumeCompleteMessage, where it's used to refer to the fragmentCallback function of the Server.

jimmyt857:
I'm also wondering if it would bother you to make a parent class Agent for both Server and Client. I wouldn't give it any virtual functions, so there would be no runtime polymorphism. Without this, I can't think of a way to refer to socketData->agent->fragmentCallback() where socketData->agent is simply a templated pointer. Without the shared functionality I would need to make 2 copies of many functions, at which point we might as well not use templates at all.

alexhultman:
In any case you need to move `"127.0.0.1", port, true` to the connect function. One "Client" needs to be shared over all of its sockets. So you need to be able to connect to multiple different addresses with the same "Client".

You can do whatever you want, but I don't want any runtime penalties. Use templates, inlines etc. You can also use virtual functions, they are zero cost when the type of the object is being known at compile time (not being a pointer to a base class that is).

alexhultman:
I made this base64 generator:

```c++
#include <iostream>
using namespace std;

int main()
{
    srand(time(0));
    char palette[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
    static char buf[16];
    for (int i = 0; i < 16; i++) {
        buf[i] = palette[rand() % 64];
        cout << buf[i];
    }
    cout << endl;
}
```

jimmyt857:
My idea was to have a Client represent a single connection. I can still share the buffers by making them static thread_local. But maybe you're right, it will be easier to just let the Client object represent many connections. The only reason I have a separate connect() function currently is that there is 1 branch of the connection process which would immediately return an error, so if I ran connect() from the constructor the user would have no time top specify the onConnectionFailure callback.

alexhultman:
You shouldn't have callbacks registered to every connection. That's how ws does it and thus the wrong way of doing it. You only need one shared object with callbacks registered and then the Socket gets passed as argument. Also, regarding the template thing, you can have templates in source files, no need to have them in headers and you should never duplicate code (except for our beginning of course). Split the functions up into smaller chunks and glue them together with inline pieces or something or use templates or anything really.

jimmyt857:
Ok, got it. And thanks for the base64 function. I'll see what I can do with all of this.

PR #148, opened 2016-06-09
Client integration

jimmyt857:
Ok, here is the merged library. I've split out the shared functionality into a base Agent class which is templated on a bool value IsServer. The compiler should be able to optimize out all of the if statements based on this, so I don't think there should be any runtime penalties.

Pretty much every class needs to be templated on the same value, which means that the user-specified callbacks now take ServerSocket and ClientSocket (typedefs of Socket<IsServer>). If you want to maintain complete backwards compatibility, we could go with AgentSocket<IsServer> and typedef Socket / ClientSocket instead.

I tried to run your benchmarks but I couldn't get the same results even before I made any changes, so definitely run them again to be sure I didn't mess something up. I'm running on an Amazon EC2 small instance so I think I just dont have the same cpu resources.


alexhultman:
I added you as a collaborator, you can merge this to client branch (doesn't exist yet). You need to fix the indentation with clang-format or anything like that and you could test it out with Autobahn for clients.

alexhultman:
It still seems to have buffers per Client, that needs to change. Some kind of shared object for holding shared resources. You can consider the sources a prototype, that is do whatever you want as long as it doesn't affect performance. I need to refactor things anyways.

jimmyt857:
Each Client object only has a single send/recv/etc buffer, and it represents many connections. Is that not sharing? The buffers and every instance of using them are copied directly from the Server, so I'm not sure what the difference is.

alexhultman:
Right..

alexhultman:
Brain fart

PR #169, opened 2016-06-28
onPing listener

sanderploegsma:
In our case (and I'm sure a lot of other cases as well) it would be more logical to have clients ping the server and handle those accordingly. To support this, I added an onPing event listener. When a `ping` is received, a `pong` is still sent back, but a callback is also triggered.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -51,6 +51,10 @@ int main()
             }*/
         });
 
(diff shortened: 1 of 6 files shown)


alexhultman:
Nice :thumbsup:

PR #175, opened 2016-07-04
Add CMake dependency lookup, force use C++11 standard.

chemhack:
find_package is used to look for openssl and zlib.
find_path/find_library are used to look for libuv, as CMake do not have built-in support for libuv lookup.

This change make cmake finding all decencies automatically and report any missing library at configure time instead of compiling or linking time.

CMAKE_CXX_STANDARD is also set to force use C++11. target_compile_features with cxx_nullptr statements are removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,17 +1,27 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.1)
 set(CMAKE_BUILD_TYPE Release)
+set(CMAKE_CXX_STANDARD 11)
 project (µWebSockets)
 
-add_subdirectory(examples)
+find_package(OpenSSL REQUIRED)
+find_package(ZLIB REQUIRED)
+
+find_path(LIBUV_INCLUDE_DIR uv.h)
+find_library(LIBUV_LIBRARY NAMES uv uv1)
 
 add_library(uWS SHARED src/Extensions.cpp src/HTTPSocket.cpp src/Network.cpp src/Server.cpp src/UTF8.cpp src/WebSocket.cpp)
 target_include_directories(uWS PUBLIC src)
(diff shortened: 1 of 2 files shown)


alexhultman:
This does not work for me.

```
CMake Error at CMakeLists.txt:6 (find_package):
  By not providing "Findopenssl.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "openssl", but
  CMake did not find one.

  Could not find a package configuration file provided by "openssl" with any
  of the following names:

    opensslConfig.cmake
    openssl-config.cmake

  Add the installation prefix of "openssl" to CMAKE_PREFIX_PATH or set
  "openssl_DIR" to a directory containing one of the above files.  If
  "openssl" provides a separate development package or SDK, be sure it has
  been installed.


-- Configuring incomplete, errors occurred!
See also "/home/alexhultman/uWebSockets/CMakeFiles/CMakeOutput.log".
```
Are you having trouble with the current CMake project? I want to use as little of CMake as possible as my experiences with CMake has been 95% negative. It never works. Never ever. CMake is a pile of rotten garbage IMO. The only reason I use it is because of Windows.

chemhack:
@alexhultman thank you for the quick feedback. I'm on a case insensitive filesystem and didn't realize CMake's find_package need strict `OpenSSL` and `ZLIB` as parameters. I updated the commit.

I agree your opinion with CMake in general, but having to provide paths in environment variables has been more pain than CMake.

alexhultman:
This works for me but it doesn't solve any problem (for me). Does it solve a problem for you? Are you on Windows? I can merge it if it solves a problem, but not if it is purely for the sake of using CMake features.

chemhack:
Yes, it solves problem for me on macOS with dependencies installed by homebrew. Without this, Makefile generated by CMake can not compile successfully.

alexhultman:
Awesome then. Thanks

PR #177, opened 2016-07-05
Support certificate chain file, and set SSL_MODE_RELEASE_BUFFERS.

chemhack:
Use `SSL_CTX_use_certificate_chain_file` instead of `SSL_CTX_use_certificate_file` allows us use certificate issued by an intermediate CA(which is the case for most of production certificates).

SSL_MODE_RELEASE_BUFFERS will greatly reduce memory consumption per connection, from our test, memory usage per idle wss connection dropped from 20KB to 10KB, which allows us to host 2x more clients with same server.


alexhultman:
Could you update the PR against master and also remane certFileName -> certChainFileName in header and cpp?

chemhack:
updated.

src/Server.cpp
```diff
@@ -310,7 +310,7 @@ SSLContext::SSLContext(std::string certFileName, std::string keyFileName)
 
     SSL_CTX_set_options(sslContext, SSL_OP_NO_SSLv3);
 
-    if (SSL_CTX_use_certificate_file(sslContext, certFileName.c_str(), SSL_FILETYPE_PEM) != 1) {
+    if (SSL_CTX_use_certificate_chain_file(sslContext, certChainFileName.c_str()) < 0) {
```
alexhultman: This needs to check != 1

chemhack: Just saw your last commit. Yeah, OpenSSL is full of consistency, :trollface:

alexhultman:
:+1:

PR #187, opened 2016-07-13
Send client verification failure reasons to socket

sanderploegsma:
Kind of speaks for itself. I wasn't sure what to use as default messages, so I'm curious what you think. Feel free to edit them as you like. 

Also, `socket.end()` is replaced by `socket.destroy()`. Not sure which of them should be used. `ws` uses destroy, although it should be used in case of failure (which this isn't). Any thoughts?

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -4,6 +4,9 @@ const http = require('http');
 const EventEmitter = require('events');
 const EE_ERROR = "Registering more than one listener to a WebSocket is not supported.";
 function noop() {}
+function abortConnection(socket, code, name) {
+    socket.end('HTTP/1.1 ' + code + ' ' + name + '\r\n\r\n');
+}
 const uws = (() => {
     try {
         return require(`./uws_${process.platform}_${process.versions.modules}`);
@@ -316,10 +319,7 @@ class Server extends EventEmitter {
                                         this.emit('connection', ws);
                                     });
                                 } else {
-                                    // todo: send code & message
-                                    if (this._lastUpgradeListener) {
-                                        socket.end();
-                                    }
+                                    abortConnection(socket, code, name);
                                 }
                             });
                         } else {
@@ -328,10 +328,7 @@ class Server extends EventEmitter {
                                     this.emit('connection', ws);
                                 });
                             } else {
-                                // todo: send code & message
-                                if (this._lastUpgradeListener) {
-                                    socket.end();
-                                }
+                                abortConnection(socket, 400, 'Client verification failed');
                             }
                         }
                     } else {
@@ -341,7 +338,7 @@ class Server extends EventEmitter {
                     }
                 } else {
                     if (this._lastUpgradeListener) {
-                        socket.end();
+                        abortConnection(socket, 400, 'URL not supported');
                     }
                 }
             }));


nodejs/dist/uws.js
```diff
@@ -318,7 +332,7 @@ class Server extends EventEmitter {
                                 } else {
                                     // todo: send code & message
                                     if (this._lastUpgradeListener) {
```
alexhultman: This is my bad, we should always close a socket that matches our path but fails the verification (only we have this path so we need to close it in all cases)

nodejs/dist/uws.js
```diff
@@ -289,6 +289,20 @@ class Server extends EventEmitter {
...
+            const response = [
+              'HTTP/1.1 ' + code + ' ' + name,
+              'Content-Type: text/html'
+            ];
+            socket.write(response.concat('', '').join('\r\n'));
+          }
+          catch (e) { /* ignore errors - we've aborted this connection */ }
+          finally {
+            // ensure that an early aborted connection is shut down completely
+            try { socket.destroy(); } catch (e) {}
```
alexhultman: This calls close(fd), so it won't guarantee what you send will be delivered as opposed to end which calls shutdown.

nodejs/dist/uws.js
```diff
@@ -289,6 +289,20 @@ class Server extends EventEmitter {
         this._upgradeListener = null;
         this._noDelay = options.noDelay === undefined ? true : options.noDelay;
         this._lastUpgradeListener = true;
+        this._abortConnection = (socket, code, name) => {
```
alexhultman: https://nodejs.org/api/http.html#http_response_writehead_statuscode_statusmessage_headers

writeHead formats the status stuff, then you just call end() to shutdown.

alexhultman:
Basically you could remove _abortConnection and just use response.writeHead followed by response.end(). And you can also remove the two top _lastUpgradeListener checks.

nodejs/dist/uws.js
sanderploegsma: Right, but there is no response around, only the socket. Or am I missing something?

sanderploegsma:
Great, thanks for the feedback! I'll get back to you!

nodejs/dist/uws.js
alexhultman: I see. I guess one have to manually construct the HTTP response then.

sanderploegsma:
Right, so since there is no `response` object to work with, I left the `_abortConnection`, just replaced `socket.destroy()` with `socket.end()`. That should do it.

nodejs/dist/uws.js
```diff
@@ -341,7 +349,7 @@ class Server extends EventEmitter {
                     }
                 } else {
                     if (this._lastUpgradeListener) {
-                        socket.end();
+                        this._abortConnection(socket, 400, 'URL not supported')
```
alexhultman: semicolon

nodejs/dist/uws.js
```diff
@@ -289,6 +289,20 @@ class Server extends EventEmitter {
...
+              'HTTP/1.1 ' + code + ' ' + name,
+              'Content-Type: text/html'
+            ];
+            socket.write(response.concat('', '').join('\r\n'));
+          }
+          catch (e) { /* ignore errors - we've aborted this connection */ }
+          finally {
+            // ensure that an early aborted connection is shut down completely
+            try { socket.end(); } catch (e) {}
+          }
```
alexhultman: indentation (2 space vs 4 space)

nodejs/dist/uws.js
```diff
@@ -289,6 +289,20 @@ class Server extends EventEmitter {
         this._upgradeListener = null;
         this._noDelay = options.noDelay === undefined ? true : options.noDelay;
         this._lastUpgradeListener = true;
+        this._abortConnection = (socket, code, name) => {
+          try {
+            const response = [
+              'HTTP/1.1 ' + code + ' ' + name,
+              'Content-Type: text/html'
+            ];
+            socket.write(response.concat('', '').join('\r\n'));
```
alexhultman: Can't this be made much simpler, something like

`socket.end('HTTP/1.1 ' + code + ' ' + name + '\r\n\r\n');`

alexhultman: The abortConnection function doesn't have to be attached to the Server, it's just a regular function that only depends on its arguments so it can be private to the module (by not adding it to this).

alexhultman:
```javascript
function abortConnection(socket, code, name) {
    socket.end('HTTP/1.1 ' + code + ' ' + name + '\r\n\r\n');
}
```

alexhultman:
I guess this solves it? I don't really know if socket.end can throw. Maybe it can, but that can be fixed later on.

sanderploegsma:
If it can, that would have been an issue before the changes as well I guess. We'll see. Thanks for merging!

PR #204, opened 2016-07-30
Add support for multiple listeners

Joncom:
- `ws` already supports this
- multiple listeners can be very useful
- this makes it easier to switch from `ws` to `uws`

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 * Permessage-deflate built-in.
 * Node.js binding exposed as the well-known `ws` interface.
 * 10-300x faster than `ws` (if they are "fastest", we are "fastester").
-* Optional engine in projects like Socket.IO, Primus & SocketCluster.
+* Default engine in SocketCluster & deepstream.io, optional in Socket.IO & Primus.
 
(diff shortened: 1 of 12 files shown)


alexhultman:
Thanks for interest.

This change would mean a massive performance and memory overhead hit that I'm not going to accept. The ws interface is already a massive mountain of bloat where each connection has its own set of callbacks and events are triggered via string arguments.

A more sane approach is how libwebsockets and WebSocket++ does it where you only have one callback for all sockets (this is also how µWS does it in C++ land). The only reason I implement the ws interface (or at least somewhat) is because of the massive hype ws comes with and its wide user base.

As you can bechmark, the ws interface adds about 5x performance overhead to the C++ interface and comes with something close to 8x memory overhead. This is because of how insanely bad the ws interface was designed in terms of scalability and performance.

lpinca:
@alexhultman already stated that he doesn't want this feature but I'm sure it will be reproposed in future. If this is ever going to be considered please make the `Socket` inherit from a real `EventEmitter` and remove the current mess.

alexhultman:
uws will never be a 100% identical drop-in replacement for ws, just look at issue #205. Some things are just too insane to implement.

PR #221, opened 2016-08-25
add minimum version requirements for dependencies

agauniyal:
for zlib, openssl, libuv dependencies under meson build

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,9 +3,9 @@ project('µWebSockets', 'cpp', default_options : ['cpp_std=c++11'])
 inc = include_directories('src')
 
 thread_dep = dependency('threads')
-zlib_dep = dependency('zlib')
-ssl_dep = dependency('openssl')
-uv_dep = dependency('libuv')
+zlib_dep = dependency('zlib', version : '>=1.0')
+ssl_dep = dependency('openssl', version : '>=1.0.0')
+uv_dep = dependency('libuv', version : '>=1.0')
 
 subdir('src')
 subdir('examples')


agauniyal:
as noted in README.md

>Dependencies

>First of all you need to install the required dependencies. On Unix systems this is typically done via package managers, like homebrew in the case of OS X or dnf in the case of Fedora Linux. On Windows you need to search the web for pre-compiled binaries or simply compile the dependencies yourself.

>libuv 1.x
OpenSSL 1.0.x
zlib 1.x
CMake 3.x

alexhultman:
Is this needed? I don't think these version numbers are correct, I actually do no know. Your check "greater than or equal" 1.0.0 seems wrong.

agauniyal:
This isn't needed necessarily but meson can show what fails the build as per as required versions are concerned. For example, when I installed libuv-dev (0.x branch) it gave warnings for versions below 1.x, and hence I installed libuv1-dev to solve it. You can close the PR if you want, this is just a suggestion 😄 and not a requirement.

socmag:
Hey I don't know if you did this already Abhinav,  but probably would be
good to have the samples build with meson as well if you are going to add
it to the lib.

agauniyal:
@socmag which samples are you referring to? These libraries - zlib, openssl, libuv?

socmag:
The uWebSocket samples

agauniyal:
echo and multithreaded echo (the samples) are being built too. Also if you want to integrate uWS into your project via meson, you just need to do - 

- add a file under subprojects directory - `uWS.wrap` with following content
```
[wrap-git]
directory=uWebSockets
url=https://github.com/uWebSockets/uWebSockets.git
revision=head
```

- add following to your main `meson.build`
```
uWS_dep = dependency('uWS', required : false)
if uWS_dep.found()
  uWS_inc = []
  uWS_lib = []
else
  uWS_sp = subproject('uWS')
  uWS_inc = uWS_sp.get_variable('inc')
  uWS_lib = uWS_sp.get_variable('uWSlib')
endif
```

You now have access to `uWS_inc` which are header file includes and `uWS_lib` which is library built for uWS which you can link against like this -
```
executable('mainexe', 'main.cpp',
		include_directories : [uWS_inc],
		link_with : [uWS_lib])
```

socmag:
Aha, great. I was having some issues with meson as a n00b, so seeing this
definitely helps. Thanks!

PR #234, opened 2016-09-02
provide a more helpful error on Windows when loading fails

silverwind:
This provides the user with a lead to the actual cause instead of wrongfully blaming the build environment 😉.

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -2,7 +2,7 @@
 
 const http = require('http');
 const EventEmitter = require('events');
-const EE_ERROR = "Registering more than one listener to a WebSocket is not supported.";
+const EE_ERROR = 'Registering more than one listener to a WebSocket is not supported.';
 function noop() {}
 function abortConnection(socket, code, name) {
     socket.end('HTTP/1.1 ' + code + ' ' + name + '\r\n\r\n');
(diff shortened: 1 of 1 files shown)


agauniyal:
single quote strings?

alexhultman:
This change is needed but it's not correct like this. Let's say you are on Node.js 6.5.0 x86 with no compiler installed. Then you will get a message to update to 6.4.0. Also, it will display that you should update to 6.4.0 even if you are on 7.3.0.

A simpler change is to just explicitly fail on win32 && version < 6.4.0 and in all other cases show the old message. I'm not sure the require will even fail on 6.3.0.

silverwind:
I'll fix the wrong version comparison logic.

> I'm not sure the require will even fail on 6.3.0

It does and it was the reason I submitted this PR.

> Let's say you are on Node.js 6.5.0 x86 with no compiler installed. Then you will get a message to update to 6.4.0.

I don't quite follow. I haven't tested it, but I assume the build step will fail in this case. Of course we could also append a message about a possible required compiler in addition to 6.4.0.

alexhultman:
Ah, my bad.

nodejs/dist/uws.js
```diff
@@ -11,8 +11,17 @@ const uws = (() => {
...
     } catch (e) {
-        throw new Error('Compilation of µWebSockets has failed and there is no pre-compiled binary ' +
-        'available for your system. Please install a supported C++11 compiler and reinstall the module \'uws\'.');
+        const version = process.version.substring(1).split('.').map(function(n) {
+            return parseInt(n);
+        });
+        const lessThanSixFour = version[0] < 6 || (version[0] === 6 && version[1] < 4);
+        if (process.platform === 'win32' & lessThanSixFour) {
+            throw new Error('µWebSockets requires Node.js 6.4.0 or greater on Windows. ' +
+            ' Please update!');
```
alexhultman: What's up with this double space?

alexhultman:
Hey btw, is it okay if I add this Droppy project to the show-and-tell? :stuck_out_tongue_winking_eye:

silverwind:
Double space fixed and force-pushed.

silverwind:
@alexhultman I'd love if you do :)

alexhultman:
:+1:

PR #271, opened 2016-10-05
Fix ws.ping

### LINKED ISSUE #270: ws.ping throws `ReferenceError: send is not defined`
```
node_modules/uws/uws.js:228
        send(message, WebSocketClient.OPCODE_PING);
        ^

ReferenceError: send is not defined
    at WebSocket.ping (node_modules/uws/uws.js:228:9)
```

silverwind:
fixes #270

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -225,7 +225,7 @@ class WebSocket {
     }
 
     ping(message, options, dontFailWhenClosed) {
-        send(message, WebSocketClient.OPCODE_PING);
+        this.send(message, WebSocketClient.OPCODE_PING);
     }
 
     // from here down, functions are not common between client and server


alexhultman:
Will this call the correct version of send, depending on what "this" is? As you can see, send differs between WebSocket and WebSocketClient. Will that work, or will the WebSocketClient act as a WebSocket when it calls into ping?

SEAPUNK:
I'm pretty sure it will call `ping` of `WebSocketClient`. So `super.send()` instead?

silverwind:
Hmm, I see that it sends garbage at least (opcode 2). super.send gives 

```
TypeError: (intermediate value).send is not a function
```

SEAPUNK:
Yeah, `super` invokes the extended class's method, but since the method is part of the extended class itself, I'm not sure where `this` plays a part.

alexhultman:
If you look at the stuff I added to the on function it seems logical that "this.send" should call the correct version of send?

SEAPUNK:
Can you try `native.server.send(this.external, message, OPCODE_PING`?

silverwind:
Yep, that works for the server, I see the ping frames. But will it do the correct thing for the client?
```
native.server.send(this.external, message, WebSocketClient.OPCODE_PING);
```

SEAPUNK:
For client, it should be a `pong()` method.

alexhultman:
The client should call native.client.ping

silverwind:
IIRC, both client and server can initate both `ping` and `pong`.

alexhultman:
Pong is not something you send from the application level, that is part of the library and the standard (both server and client automatically respond to pings with a pong).

SEAPUNK:
Yeah. I just misunderstood what silverwind meant by the original comment. I missed that there's both a `native.client` and a `native.server`.

silverwind:
How would I discern a call from a client to one from a server?

silverwind:
Feel free to take over ;)

SEAPUNK:
**OH OH OH**, okay I see what the WebSocketClient class is doing.

So in this case, I'm pretty sure you just create a `ping()` method on the Client class to "shadow" the `WebSocket` class's method, calling `native.client.send`.

alexhultman:
I did test this by printing in the both send() functions and ping() does call the correct send() with this PR applied.

alexhultman:
However, the actual implementation of ping is completely wrong.

alexhultman:
It has to be done like this: 7baccde2616ffcd27388bd22167c4e9b0d42c6ea

alexhultman:
Can you grab the uws.js of master and just overwrite your version and test it. If it works I will push a new version.

alexhultman:
Never mind, it works now. I published a new version.

silverwind:
Looking good from my side:

silverwind:
Btw, I'd be great if you could move `package.json` to the root of the project, so one can install a unreleased version with just `npm install git@github.com:uWebSockets/uWebSockets.git`.

alexhultman:
I'm off to bed, will check on that later. Thanks.

PR #303, opened 2016-10-19
Small build tweaks: Allow static, enable Debian stable, libuv location

not-implemented:
While integrating µWebSockets into a standalone C++-server as a submodule, we needed to make some changes, which would be nice to get them integrated into upstream:

- Enable build with CMake 3.0: CMake 3.0 does not implement "CMAKE_CXX_STANDARD" - polyfill for GCC added to make compiling on current Debian stable possible
- Allow build of static library:  Useful when including directly from a parent CMake project. I don't know, if setting "BUILD_SHARED_LIBS" to "ON" globally is best practice, but I didn't find a better solution
- Allow passing libuv location (from a parent module): If any parent module uses its own libuv dependency, we want to also use this one
- Updated .gitignore: When using µWebSockets as GIT-Submodule, all unversioned files causes the parent repo to be marked as modified

I hope these changes does not break anything somewhere else :-) Let me know, if you have better solutions for this.


alexhultman:
Sorry for delay. I honestly don't care at all about CMake so I'll just trust you this works.

not-implemented:
Delay?! Was just 2 days, which is really fast ;-) Thanks.

Issue #308, opened 2016-10-21
FD_CLOEXEC should be add to (socket)


codehz:
`uv_spawn()` not check if all other fd have been closed, so the sub-process will inheritance those fd opend by uWebSockets....

PR #309, opened 2016-10-21
add FD_CLOEXEC after create socket

codehz:
#308 Fixed


alexhultman:
Could you fix the style and make it wirk on Windows and maybe wrap rhe call to socket. You could have a wrapper called privateSocket that calls socket and sets flags on unix but not on Windows

codehz:
@alexhultman I have fixed it

alexhultman:
We also need to set NO_DELAY so maybe it is better to just have a simple function taking and returning fd while setting all the common default flags.

alexhultman:
Fd = setDefaultFlags(fd)

alexhultman:
Alao - this PR does not follow the style of the project

PR #318, opened 2016-10-31
Properly handle error codes in the Private Use range (4000-4999)

### LINKED ISSUE #317: CloseFrame parsing rejects valid error codes
I noticed after switching a client to uws from ws that Private Use error codes were no longer properly handled. Status Codes 4000-4999 should always be valid.

deckerrj:
Fixes #317

### DIFF
--- a/src/WebSocketProtocol.h
+++ b/src/WebSocketProtocol.h
@@ -246,8 +246,8 @@ public:
         if (length >= 2) {
             memcpy(&cf.code, src, 2);
             cf = {ntohs(cf.code), src + 2, length - 2};
-            if (cf.code < 1000 || cf.code > 1011 || (cf.code >= 1004 && cf.code <= 1006) ||
-                !isValidUtf8((unsigned char *) cf.message, cf.length)) {
+            if (cf.code < 1000 || cf.code > 4999 || (cf.code > 1011 && cf.code < 4000) ||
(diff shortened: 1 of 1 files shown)


alexhultman:
How would you interpret this:

```
   3000-3999

      Status codes in the range 3000-3999 are reserved for use by
      libraries, frameworks, and applications.  These status codes are
      registered directly with IANA.  The interpretation of these codes
      is undefined by this protocol.

   4000-4999

      Status codes in the range 4000-4999 are reserved for private use
      and thus can't be registered.  Such codes can be used by prior
      agreements between WebSocket applications.  The interpretation of
      these codes is undefined by this protocol.
```
Question is if 3000-4999 should be allowed?

alexhultman:
If they are undefined by this protocol, they should probably not be force closed.

deckerrj:
It seems like 3000-3999 are basically reserved for future registration of libraries with the IANA. Looks like they are currently unassigned (https://www.iana.org/assignments/websocket/websocket.xml#close-code-number) and therefore would not be valid close codes yet.

alexhultman:
My interpretation is still that 3000-4999 should all be ignored (aka pass with no close). When (if) anone registers say code 3000 we shouldn't need to update the lib.

alexhultman:
Lets merge for now

PR #322, opened 2016-11-01
Pass the Sec-Websocket-Protocol header through from Node

BenV:
This PR allows the `Sec-Websocket-Protocol` header to be passed from Node to C++ so that the header can be echoed back in the response. This prevents the following error: `Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received`, similar to https://github.com/uWebSockets/uWebSockets/issues/263

Please let me know if there's anything you'd like done differently, as I am very new to this repo. Thanks!


alexhultman:
Perfect

PR #332, opened 2016-11-16
add support for libuv >= 2.0.0

### LINKED ISSUE #329: Does not compile against libUV 2.x
In livUV 2.x we have version tags like
```
#define UV_VERSION_MAJOR 2
#define UV_VERSION_MINOR 0
#define UV_VERSION_PATCH 0
```
https://github.com/libuv/libuv/blob/master/include/uv-version.h

But Hub.h only check minor version, which does not accept the 2.x version
```
static_assert (UV_VERSION_MINOR >= 3, "µWebSockets requires libuv >=1.3.0");
```

lucasterra:
libuv [master branch is 2.0.0](https://github.com/libuv/libuv/blob/master/include/uv-version.h). Closes #329

### DIFF
--- a/src/Hub.h
+++ b/src/Hub.h
@@ -7,7 +7,7 @@
 #include <zlib.h>
 #include <mutex>
 
-static_assert (UV_VERSION_MINOR >= 3, "µWebSockets requires libuv >=1.3.0");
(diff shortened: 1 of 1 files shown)


SEAPUNK:
Sure, the master branch is 2.0, but IIRC, they have not released 2.0 yet, so we have no idea what their API is, or what breaking changes we need to look out for.

alexhultman:
I would merge this for now though (it *might* work in 2.0 with no changes except for this one).

PR #336, opened 2016-11-20
allow access to httpheaders on connection handler

lucasterra:
This exposes the HTTPParser to onConnection's upgradeinfo. I have also implemented a simpler way to parse the headers, through `uWS::UpgradeInfo::getRequestHeaders()`, which returns a `std::map<std::string, std::string>`. There's a small overhead of copying/allocating the strings and the map, but people can still use the HTTPParser if zero-copy parsing is required.

**Example:**
```cpp
int main()
{
    uWS::Hub h;

    h.onConnection([](uWS::WebSocket<uWS::SERVER> ws, uWS::UpgradeInfo ui) {
        // headers
        const std::map<std::string, std::string>& headers = ui.getRequestHeaders();
        for (auto i = headers.begin(); i != headers.end(); i++) {
            // i->first: header key; i->second: header value
            std::cout << i->first << ": " << i->second << std::endl;
        }
    });

    h.onMessage([](uWS::WebSocket<uWS::SERVER> ws, char *message, size_t length, uWS::OpCode opCode) {
        ws.send(message, length, opCode);
    });

    h.listen(3000);
    h.run();
}
```

### DIFF
--- a/src/Group.h
+++ b/src/Group.h
@@ -15,10 +15,10 @@ struct Hub;
 struct UpgradeInfo {
     char *path, *subprotocol;
     size_t pathLength, subprotocolLength;
-    uWS::HTTPParser httpParser;
+    HTTPParser httpParser;
 
     std::map<std::string, std::string> getRequestHeaders() {
-        char* curCursor = httpParser.cursor;
+        char *curCursor = httpParser.cursor;
         httpParser.cursor = httpParser.headerBegin;
         std::map<std::string, std::string> headers;
 
--- a/src/Group.h
+++ b/src/Group.h
@@ -14,32 +14,32 @@ struct Hub;
 
 struct UpgradeInfo {
     char *path, *subprotocol;
-	size_t pathLength, subprotocolLength;
-	uWS::HTTPParser httpParser;
+    size_t pathLength, subprotocolLength;
+    uWS::HTTPParser httpParser;
 
-	std::map<std::string, std::string> getRequestHeaders() {
-		char* curCursor = httpParser.cursor;
-		httpParser.cursor = httpParser.headerBegin;
-		std::map<std::string, std::string> headers;
+    std::map<std::string, std::string> getRequestHeaders() {
+        char* curCursor = httpParser.cursor;
+        httpParser.cursor = httpParser.headerBegin;
+        std::map<std::string, std::string> headers;
 
-		for (httpParser++; httpParser.key.second; httpParser++) {
-			std::string key;
-			key.resize(httpParser.key.second);
+        for (httpParser++; httpParser.key.second; httpParser++) {
+            std::string key;
+            key.resize(httpParser.key.second);
 
-			size_t i = 0;
-			for (; i < httpParser.key.second; i++) {
-				key[i] = tolower(httpParser.key.first[i]);
-			}
-			key[i] = 0;
+            size_t i = 0;
+            for (; i < httpParser.key.second; i++) {
+                key[i] = tolower(httpParser.key.first[i]);
+            }
+            key[i] = 0;
 
-			headers[key] = std::string(httpParser.value.first, httpParser.value.second);
-		}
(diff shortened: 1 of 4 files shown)


alexhultman:
Could you replace the tabs with 4 spaces?

lucasterra:
Sure! Will resubmit it in a few minutes.
Em dom, 20 de nov de 2016 às 20:26, Alex Hultman <notifications@github.com>
escreveu:

lucasterra:
Done.

lucasterra:
Uhh... wait

lucasterra:
My Visual Studio overwrote something wrongly by mistake. I've setup indentation rules for this specific project. I will match the project style for the next PRs. 👍

src/Group.h
```diff
@@ -13,6 +15,31 @@ struct Hub;
...
+
+        for (httpParser++; httpParser.key.second; httpParser++) {
+            std::string key;
+            key.resize(httpParser.key.second);
+
+            size_t i = 0;
+            for (; i < httpParser.key.second; i++) {
+                key[i] = tolower(httpParser.key.first[i]);
+            }
+            key[i] = 0;
```
alexhultman: This stores 0 out of bounds (and it is not needed).

You can use std::transform(key.begin(), key.end(), key.begin(), ::tolower); instead since you already have the key as a std::string.

So you could do like this:
std::string key(httpParser.key.first, httpParser.key.second);
std::transform(key.begin(), key.end(), key.begin(), ::tolower);
headers[key] = ...

src/Group.h
```diff
@@ -13,6 +15,31 @@ struct Hub;
 struct UpgradeInfo {
     char *path, *subprotocol;
     size_t pathLength, subprotocolLength;
+    uWS::HTTPParser httpParser;
```
alexhultman: Don't need uWS:: here

src/Group.h
```diff
@@ -13,6 +15,31 @@ struct Hub;
...
+
+    std::map<std::string, std::string> getRequestHeaders() {
+        char* curCursor = httpParser.cursor;
```
alexhultman: star to the right

src/Group.h
alexhultman: Oh wait, I see you skip one copy. Okay but you need to remove the 0

lucasterra:
Pushed the coding style issues and the array out of bounds thing.
`std::transform` is not really needed, we can make it in one pass. Unless you think it's gonna make it more readable, or something...

src/HTTPSocket.h
```diff
@@ -34,6 +34,42 @@ struct WIN32_EXPORT HTTPSocket : private uS::Socket {
     static void onEnd(uS::Socket s);
 };
 
+struct HTTPParser {
+    char* headerBegin;
```
alexhultman: Lol these are to the left

src/Group.h
```diff
@@ -13,6 +15,31 @@ struct Hub;
...
+        char *curCursor = httpParser.cursor;
+        httpParser.cursor = httpParser.headerBegin;
+        std::map<std::string, std::string> headers;
+
+        for (httpParser++; httpParser.key.second; httpParser++) {
+            std::string key;
+            key.resize(httpParser.key.second);
+
+            // lowercase the header key
+            size_t i = 0;
```
alexhultman: put it in the for?

alexhultman:
These are pedantic comments but I guess I need to enforce a common style to not get things messed up

lucasterra:
Yeah, I understand your concern. But it's not exactly easy to adapt to someone's code style.

alexhultman:
Thanks, I will merge this locally and fix some small pedantic things and amend commit that

lucasterra:
All right. Have fun! 😃

lucasterra:
Where did this go?

alexhultman:
I reset the merge since the server started to segfault with that commit. Will have to be reimplemented sometime.

Issue #349, opened 2016-11-28
Node-gyp rebuild fails on Windows (cmd/powershell)


LsKs:
Include the following information (where applicable):

* What operating system are you running?
  * Windows
* What platform/language are you using?
  * Node.js
* Are you using SSL, if so, how?
  * No
* Can you show me some short code that triggers your issue?

When installing the module from NPM on Node v6.9.1 (32-bit in this case), `node-gyp rebuild` fails to move the binary to the correct location as it refers to `cp` which obviously isn't available in cmd.exe (triggering in powershell doesn't work as well). Installing from cygwin works as expected.

Line 57 of binding.gyp:
`'action': ['cp', '<@(PRODUCT_DIR)/uws.node', 'uws_<!@(node -p process.platform)_<!@(node -p process.versions.modules).node']`

Build log attached.
[build_log.txt](https://github.com/uWebSockets/uWebSockets/files/616960/build_log.txt)

alexhultman:
Do you know how to easily make it call copy on Windows and cp on everything else?

LsKs:
Unfortunately I don't possess such a knowledge, still learning about C++ in Node.

PR #350, opened 2016-11-28
Fix for node-gyp rebuild on Windows (cmd/powershell)

LsKs:
As described in #349.

Apologies if it's not the right way to do it, initially I tried to use `copies` but that didn't seem to have the option to change a name.


alexhultman:
Hey it's better than nothing. Thanks

PR #358, opened 2016-12-06
Fix CMakeLists.txt to specify C++ 11 standard

qwename:
Use `set_property` instead of `set` for `CMAKE_CXX_STANDARD` or
`CXX_STANDARD`. Only tested with GNU compilers.

The compiler check can probably be omitted if the flag was changed to `--std=c++11`.


alexhultman:
Thanks, I haven't tested this but I guess it is better than what was :)

PR #361, opened 2016-12-06
added multithreaded echo example based on pre-update version

TheMungBall:


alexhultman:
:+1:

it is possibke to remove addAsync of the main hub. addAsync is like "listen for transfers from other threads"

PR #363, opened 2016-12-08
Removed unnecessary 'addAsync' from multithreaded example

TheMungBall:


alexhultman:
Your base commit is not the same as the one i merged so it conflicts

Issue #375, opened 2016-12-24
Compilation error - Fedora 24 GCC 6.2.1


ng5:
Hi,

I'm unable to compile the code due to below error. Can you please take a look?

thanks

/home/rohitv/GitHub/uWebSockets/examples/multithreaded_echo.cpp:19:59: error: ‘uWS::UpgradeInfo’ has not been declared
   h.onConnection([&](uWS::WebSocket<uWS::SERVER> ws, uWS::UpgradeInfo upgInfo){
                                                           ^~~~~~~~~~~
/home/rohitv/GitHub/uWebSockets/examples/multithreaded_echo.cpp:23:4: error: no matching function for call to ‘uWS::Hub::onConnection(main()::<lambda(uWS::WebSocket<true>, int)>)’
   });
    ^
In file included from /home/rohitv/GitHub/uWebSockets/examples/../src/Hub.h:4:0,
                 from /home/rohitv/GitHub/uWebSockets/examples/../src/uWS.h:4,
                 from /home/rohitv/GitHub/uWebSockets/examples/multithreaded_echo.cpp:1:
/home/rohitv/GitHub/uWebSockets/examples/../src/Group.h:52:10: note: candidate: void uWS::Group<isServer>::onConnection(std::function<void(uWS::WebSocket<isServer>, uWS::HTTPRequest)>) [with bool isServer = true]
     void onConnection(std::function<void(WebSocket<isServer>, HTTPRequest)> handler);
          ^~~~~~~~~~~~

qwename:
`uWS::UpgradeInfo` doesn't exist anywhere in the source, change the parameter to `uWS::HTTPRequest httpReq` to make it build.

alexhultman:
The general idea is that if something is broken -> check the latest released version.

PR #376, opened 2016-12-24
Fixed multithreading example

qwename:
See #375, uWS::UpgradeInfo doesn't exist.

EDIT: Also changed printing of messages to hexdump. and commented it as it greatly reduced echo throughput (~4x as slow).


alexhultman:
This fix is intertwined with unnecessary hex print functions so I cannot merge it. Of course you are going to kill performance by printing to stdout, the example is just showing how to move connections between threads

qwename:
I've removed the hexdump function.

PR #378, opened 2016-12-26
Fix images in npmjs

DiegoRBaquero:
Use full URL to fix image on npmjs website


alexhultman:
The question I would like to ask is why? Why would NPM suddenly break the images that has been working for a year now. All other project browsers display the images correctly but NPM. I would say this is a bug with NPM itself and they need to fix their broken site.

DiegoRBaquero:
Maybe it works using markdown (?) but using the html version to center did break it?

alexhultman:
I haven't touched any of that in a long time. They just decided to stop fetch images from github.

DiegoRBaquero:
Opened an issue: https://github.com/npm/www/issues/119

DiegoRBaquero:
Either needs to be uncentered or use absolute url for now. 

https://github.com/npm/marky-markdown/issues/320

alexhultman:
well NPM broke it and they should fix it

alexhultman:
thanks for reporting

PR #392, opened 2017-01-13
Propose to silence warnings for unused parameters

coin-au-carre:
These warnings are visible when using -Wextra flags.


alexhultman:
Hi,

Some of these silencers are actually hiding incomplete features:

* ContentType is not used in HTTP response
* prepareMessageBatch is not taking into account the excludedMessages vector

I think its better to have warnings than to silence them just for the sake of silencing them. I haven't done anything with the lib for a while, I'm currently having a little bit of a cold and I'm not feeling it right now. Master is in a pretty busted shape right now but the 0.12.0 should be fine in the mean time.

coin-au-carre:
Ok thanks for your answer. I understand, I ll close the issue then. Wishing you a speedy recovery !

PR #402, opened 2017-01-24
Fix unresolved external symbol with NodeData::asyncCallback

### LINKED ISSUE #351: example echo linking error. undefined symbol uS::NodeData::asyncCallback
Hello, everyone. I use to employ uWS on OSX, and it works very well.
Recently, I have to use uWS under Windows 10 with visual studio 2015.

I successfully compile the library with CMake and Visual Studio. But When coming to echo example, there is a wired linking error: undefined symbol "public: static void __cdecl uS::NodeData::asyncCallback(struct uv_async_s *)".

I am pretty sure the project setting was right, and there is just only this one linking error.
Any help would be grateful, thanks a lot!

MergHQ:
Closes #351 

This has been a problem for windows users. The solution was found some time ago in the issue, but was never fixed. 

Let me know if there's anything that needs to be changed.

Thanks 😄

### DIFF
--- a/src/Networking.h
+++ b/src/Networking.h
@@ -108,7 +108,7 @@ Context createContext(std::string certChainFileName, std::string keyFileName, st
 
 struct SocketData;
 
-struct NodeData {
+struct WIN32_EXPORT NodeData {
     char *recvBufferMemoryBlock;
     char *recvBuffer;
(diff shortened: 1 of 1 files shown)


alexhultman:
Thanks :+1:

Issue #404, opened 2017-01-24
Makefile | Environment variable to set install location and to build static target


TheMegaTB:
It would be nice to have an environment variable or argument for the Makefile/make command to replace `/usr/lib` and `/usr/include` since some use-cases require static links and copying the files into `/usr/lib` requires (most likely) root permission.

alexhultman:
I'm focusing on a slim Makefile that should cover most use cases

PR #405, opened 2017-01-24
Added static library build target

TheMegaTB:
By default the Makefile now builds a static library target in the main directory. The .o files have been added to the clean target.
Related to issue #404

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,12 @@ default:
 	make `(uname -s)`
 Linux:
 	g++ $(CPP_SHARED) -s -o libuWS.so
+	g++ $(CPP_SHARED) -s -c
+	ar rvs libuWS.a *.o
 Darwin:
 	g++ $(CPP_SHARED) $(CPP_OSX) -o libuWS.so
+	g++ $(CPP_SHARED) $(CPP_OSX) -s -c
+	ar rvs libuWS.a *.o
 .PHONY: install
 install:
 	if [ -d "/usr/lib64" ]; then cp libuWS.so /usr/lib64/; else cp libuWS.so /usr/lib/; fi
@@ -15,3 +19,4 @@ install:
 .PHONY: clean
 clean:
 	rm libuWS.so
+	rm *.o


Makefile
```diff
@@ -5,8 +5,12 @@ default:
 	make `(uname -s)`
 Linux:
 	g++ $(CPP_SHARED) -s -o libuWS.so
+	g++ $(CPP_SHARED) -s -c
+	ar rvs libuWS.a *.o
 Darwin:
 	g++ $(CPP_SHARED) $(CPP_OSX) -o libuWS.so
+	g++ $(CPP_SHARED) $(CPP_OSX) -s -c
```
alexhultman: I'm pretty sure -s is deprecated/errors on OS X

Makefile
```diff
@@ -5,8 +5,12 @@ default:
...
+	ar rvs libuWS.a *.o
 .PHONY: install
 install:
 	if [ -d "/usr/lib64" ]; then cp libuWS.so /usr/lib64/; else cp libuWS.so /usr/lib/; fi
```
alexhultman: It should also copy the .a library when installing right

TheMegaTB:
I'm currently ill but I'll look into it probably after the weekend. Even though it results in a statically linkable file I get an awful lot of 'undefined reference' errors when attempting to use it in my CMake project so I'm going to investigate that as well.

TheMegaTB:
Okay so I attempted to use the static library target [http://pastebin.com/ATWPp98y](http://pastebin.com/ATWPp98y) but the result is the same. It apparently doesn't matter if I add only `uWS.h` or all `.h` files in the include directory.

I also attempted to link `test.cpp` against the shared library created by the Makefile and it doesn't compile as well so the problem doesn't appear to be related to the static target only [http://pastebin.com/qgSU37TB](http://pastebin.com/qgSU37TB).

TheMegaTB:
I just read through the CMakeLists.txt and it might be possible that the library is not linked against the OpenSSL and zlib

TheMegaTB:
Since the Makefile appears to be missing the OpenSSL and zlib dependency yet and I personally prefer using CMake, I created a new pull request that modified the CMakeLists.txt to produce a static target as well as making using some very handy configuration variables. Just for reference it's #417, I'll open an issue regarding the flaws in the Makefile

PR #413, opened 2017-01-28
Update uws.js - just a suggestion

ORESoftware:
This is just a suggestion. You've chosen to wrap http.Server within another class. As a suggestion, instead of doing what you did (wrapping http.Server in a class and having http.Server as a field in your class)... instead, create a function, and return a new http server from the function; I personally find that even cleaner than composition. If you really need to embellish http.Server, then extend http.Server instead of extending EventEmitter. IDK, just a thought.

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -439,9 +439,15 @@ class Server extends EventEmitter {
 
         if (options.port) {
             if (options.host) {
-                this.httpServer.listen(options.port, options.host, callback);
+                this.httpServer.listen(options.port, options.host, () => {
+                    this.emit('listening');
+                    callback();
+                });
             } else {
-                this.httpServer.listen(options.port, callback);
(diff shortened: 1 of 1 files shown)


nodejs/dist/uws.js
```diff
@@ -439,9 +439,15 @@ class Server extends EventEmitter {
 
         if (options.port) {
             if (options.host) {
-                this.httpServer.listen(options.port, options.host, callback);
+                this.httpServer.listen(options.port, options.host, () => {
```
alexhultman: Perfect, but are you sure callback never gets any arguments?

alexhultman: What happend when callback is undefined?

ORESoftware:
Correct we need to account for when cb is not defined, let me fix it, I
havent tested it

ORESoftware:
I am 99% certain that the callback takes no arguments

PR #414, opened 2017-01-28
Update uws.js

ORESoftware:
I didn't know how to update my other PR (Github does not have a perfect interface) so I had to create a new PR.
You're right the callback is not always defined of course. Also, I think the Node.js server API guarantees that if there is a callback passed, it will also be fired when the listening event fires, so this PR may not be 100% ready yet. This is just a suggestion.


alexhultman:
It looks legit to me. Maybe a style nit but who cares

PR #419, opened 2017-01-30
Add HTTP verb OPTIONS

sapeien:
Nice to have for responding to CORS pre-flight requests.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS


alexhultman:
Apparently it's also missing HEAD, TRACE and CONNECT :+1:

PR #438, opened 2017-02-06
nodejs: don't attempt to upgrade closed sockets

silverwind:
ported from: https://github.com/primus/primus/pull/531
fixes: https://github.com/uWebSockets/uWebSockets/issues/326

Not sure if the checks for `.writeable` and `.readable` like it was done for primus are necessary, so I went with a boolean check on the handle.

Can't run autobahn right now because of https://github.com/uWebSockets/uWebSockets/issues/430.

### DIFF
--- a/nodejs/dist/uws.js
+++ b/nodejs/dist/uws.js
@@ -102,7 +102,7 @@ class WebSocket {
         } else {
             this.internalOnOpen = noop;
         }
-    }   
+    }
 
     set onclose(f) {
         if (f) {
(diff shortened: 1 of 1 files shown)


nodejs/dist/uws.js
```diff
@@ -102,7 +102,7 @@ class WebSocket {
         } else {
             this.internalOnOpen = noop;
         }
-    }
```
alexhultman: :+1:

PR #450, opened 2017-02-11
Expose uv_run_mode

Oipo:
In some cases, running uws should not block the calling thread. For example, if you have another I/O process running like rdkafka and need to call both uws' run and rdkafka's run function in your own loop.

This PR simply exposes the underlying uv's option to set the mode to do just that, without breaking backwards compatibility.

### DIFF
--- a/src/Node.cpp
+++ b/src/Node.cpp
@@ -53,10 +53,10 @@ Node::Node(int recvLength, int prePadding, int postPadding, bool useDefaultLoop)
     SSL_CTX_set_options(nodeData->clientContext, SSL_OP_NO_SSLv3);
 }
 
-void Node::run(uv_run_mode mode) {
+int Node::run(uv_run_mode mode) {
     nodeData->tid = pthread_self();
 
-    uv_run(loop, mode);
+    return uv_run(loop, mode);
 }
 
 Node::~Node() {
--- a/src/Node.cpp
+++ b/src/Node.cpp
@@ -53,10 +53,10 @@ Node::Node(int recvLength, int prePadding, int postPadding, bool useDefaultLoop)
     SSL_CTX_set_options(nodeData->clientContext, SSL_OP_NO_SSLv3);
 }
 
-void Node::run() {
+void Node::run(uv_run_mode mode) {
     nodeData->tid = pthread_self();
 
-    uv_run(loop, UV_RUN_DEFAULT);
+    uv_run(loop, mode);
 }
 
 Node::~Node() {
--- a/src/Node.h
+++ b/src/Node.h
@@ -21,7 +21,7 @@ protected:
 public:
     Node(int recvLength = 1024, int prePadding = 0, int postPadding = 0, bool useDefaultLoop = false);
     ~Node();
-    void run();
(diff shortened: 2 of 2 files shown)


alexhultman:
Code looks fine but I'm not sure about your use case. If you intertwine two loops in one thread your cpu usage will be 100 and you'll get poor performance

Oipo:
Only if you don't sleep the thread yourself. The idea behind this is that you give control to the caller combined with sane defaults, instead of forcing it in the framework. As is with most things programming, shooting yourself in the foot is possible but should not be avoided at all costs.

I could of course put both loops on their own thread, but then I'd end up with having to deal with multi-threaded access to shared variables. Unless I'm mistaken, libuv should only call callbacks in the uv_run() function, meaning it would greatly simplify sharing variables between two loops.

I'd rather have to scale-out my program, than to make my program multi-threaded. Is this sufficient clarification for my PR?

alexhultman:
Wouldn't this PR be unnecessary now that we have different backends to chose from? It is a much better idea to get the project properly integrated with whatever event-loop you have rather than hack in some timer (which kills perf).

Oipo:
Hmm. While I agree with your point on needing a timer or having 100% CPU usage, neither the ASIO nor the epoll backends solve this either. I guess the only solution is to have multiple threads after all, oh well.

If you have a better suggestion, let me know. Anyways, I'm closing this.

alexhultman:
You can actually add the epoll fd of the epoll backend as a poll in another epoll set.

alexhultman:
In other words the entire event loop can be a poll in any other event loop. So we do need the ability to run only one iteration. Then you could create a poll using Loop::epfd and poll for readable and when this triggers you run one iteration of the Loop

alexhultman:
And ofc the same applies the other way around -> you can create a uWS::Poll using the epollfd of any other event-loop and instead run that other loop in single steps. It all depends on what loop you want to be primary. So you can actually solve this today very efficiently

Oipo:
If I get what you're saying, I'd have to extract the loops from the other libraries and have them be orchestrated by the master loop?

The problem here being that the other libraries I'm using currently have their own queuing systems which are not based on any of the technologies used in uws.

Oipo:
Also, another point to be taken into account: rdkafka specifically needs a semi-regular call to its loop system, at least a couple times per second. There'd probably have to be a way to generate a steady stream of wake-up events that do nothing but run the callback system at least once.

alexhultman:
All you need to get from rdkafka is an integer called the epoll file descriptor. This integer is the main part of the rdkafka event-loop and can be added to the event-loop of uws. Linux will then trigger a callback when rdkafka wants its event-loop iterated. This allows perfect integration with any event-loop on Linux systems.

alexhultman:
I can add a helper function like Loop::embedd(epfd, void(cb *)(epfd));

alexhultman:
I completely forgot about timers, they make it a lot more complex to solve..

alexhultman:
We will have to see what the best strategy is. I want both projects to integrate well, but not at the cost of performance or timers or sleeps, etc.

PR #454, opened 2017-02-13
English typos in the main README

christopherhenard:


alexhultman:
I'm no expert but "as" has a different logical meaning than "more". 2x *as* performant means 100% *more* performant, not 200% more performant. This is also why I write "equal" instead of 1x since it is not 1x more performant, it is 1x *of* the subject in perf.

daliwali:
I think @alexhultman is correct. "1x more performant" would mean it's 100% more performant, or twice as performant.

christopherhenard:
Yeap you're right.

Issue #445, opened 2017-02-09
100% cpu usage when ulimit cap is reached


jimmyt857:
Include the following information (where applicable):

* What operating system are you running?
  * Linux
* What platform/language are you using?
  * C++
* Are you using SSL, if so, how?
  * SSL via µWS

When there are enough connections to reach the linux file descriptor limit, the server begins using 100% of the cpu. The cpu usage comes from the server continually reaching this line (with errno set to EMFILE in my case):

https://github.com/uWebSockets/uWebSockets/blob/master/src/Node.h#L81

I assume this means that when accept() is unable to open a new file descriptor, it leaves the pending connection in the queue and so the listening socket continues to stay ready for reading. I'm reporting this not because I think the behavior should be changed, but because it took me awhile to track this down, so I think it would be nice to have some sort of logging (which could be turned on by an environment flag) to avoid the silent failure.

If you're happy with this solution I can make a PR. It does feel a bit silly though to add in a compiler flag for what is currently a single logging statement.

alexhultman:
A better solution would be to stop the polling of the listen fd and replace it with a 10 second timer which starts the polling again.

jimmyt857:
Haha, I actually just thought of this in the shower. But we could instead remove the listen callback from the libuv event queue and re-add it as a 50ms timer event. Whenever the accept works again it could switch back to read polling.

alexhultman:
Shower programming is the best

jimmyt857:
👍 

I'll work on this and submit a PR.

mafrost:
Nice solution. We have also encountered this with no good solution on how to handle it.

arputcu:
how can i debug/see if ulimit cap is reached?

jimmyt857:
Just add a print statement right before the return statement (the line of code in my first post) that prints errno after accept failure, then look up the errno code on google.

PR #456, opened 2017-02-13
Accept issue

jimmyt857:
This works for me. Take a look and let me know what you think. I had to copy some of the code from Socket.close() into the Group.cpp stopListening() call because there is no uv_poll_t Socket while the timer is running. There might be a way to do this a bit cleaner without the code replication.

### DIFF
--- a/src/Group.cpp
+++ b/src/Group.cpp
@@ -158,7 +158,22 @@ void Group<isServer>::stopListening() {
     if (isServer) {
         uS::ListenData *listenData = (uS::ListenData *) user;
         if (listenData) {
-            uS::Socket(listenData->listenPoll).close();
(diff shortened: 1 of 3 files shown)


jimmyt857:
#445

jimmyt857:
Whenever we're done with this, I've now also got a completely functional first draft of uUV ready to go!

alexhultman:
:+1: will look at this

alexhultman:
i trust you :+1:

jimmyt857:
Haha thanks. I'm quite confident in the stability of this, the only thing you might want to seriously consider is if there is a cleaner way to implement the Group.stopListening() changes.

Issue #446, opened 2017-02-09
Work on experimental speedups


jimmyt857:
Awhile back you described writing a toned down version of libuv (which I see is partially implemented) and also trying out mtcp. I'm interested in working on this myself, so before I start I just wanted to know if you had any design/details in mind. For example, should I use function pointers to swap in/out the mtcp calls? i.e uws_connect(), which is either set to ::connect() or mtcp_connect()

Also, I seem to recall another branch which you were using to work on this, does that have more work done than the uUV stuff in the master?

alexhultman:
I'm unemployed, I cannot afford working on ultra low level stuff nobody uses anyways. I wanted to write a complete TCP stack but it's not going to happen. I need to sell some kidneys and suck some d*ck first of all. I would not use function pointers though but hey nobody cares anyways

jimmyt857:
Haha, alright. I'll start submitting PRs on this when I can, possibly as early as this weekend.

alexhultman:
What I would like to see is a TCP stack with no per socket buffers whatsoever. The driver would just call back with the data as it comes in. Imagine having billions of connections to a single Raspberry Pi. It should be possible, source: gut feeling.

bigopon:
> but hey nobody cares anyways

I care but I don't understand why it's not preferred and pretty curious about it. Can I have some enlightenment pls?

alexhultman:
The only reason this project has more than 4 users is because it leeches off of Node.js's massive popularity. Without acting like a parasite to Node.js it would not have taken off. If you start to deviate from Node.js by fixing low level stuff that cannot possibly be used in Node.js itself you lose 99% of the interest since it doesn't apply to Nodesters anymore. In fact, even by hinting about the fact that Node.js has terrible performance leads to loss of interest since it doesn't line up with the made up unicorn chocolate land of those 99%. It is basic PR, not tech.

EduardoRFS:
We love JS(and node), and you are creating a silver bullet for us.

daliwali:
>If you start to deviate from Node.js by fixing low level stuff that cannot possibly be used in Node.js itself you lose 99% of the interest since it doesn't apply to Nodesters anymore. In fact, even by hinting about the fact that Node.js has terrible performance leads to loss of interest since it doesn't line up with the made up unicorn chocolate land of those 99%. It is basic PR, not tech.

[...]

PR #465, opened 2017-02-17
Experimental

jimmyt857:
Making this pull request not necessarily for it to be merged immediately, but to start a conversation thread. The changes affect most files because I've added a macro UWS_UV which is set either to nothing, or to "uUV::". This minimizes cases where we have to do a #ifdef USE_MICRO_UV, while also preventing name conflicts for users who use libuv in their own code.

I've tested this in my own production and it works fine (no bugs / crashes, valgrind clean). I have not actually done any benchmarking on it, though.

There is 1 significant issue left to resolve, which is that on occasion the epoll_wait will be delayed longer than usual before returning. I have not figured out what causes this yet.

One smaller question: the code you had already written used a global array of uv_poll_cb callbacks, and each uv_poll_t stores an index into that array. I kept that functionality, but is there actually any advantage to this? It seems like it adds an extra array access to every callback call without any advantage.


jimmyt857:
#446

alexhultman:
Hmmmm...

I'm not a fan of prefixing everything with UWS_UV. I'm also not a fan of introducing handle as a runtime value to every uv_handle_t. Libuv is not very efficient and it makes use of runtime dynamic dispatches based on this handle->type thing which is due to the fact C lacks function overloading. They implement function overloading in the runtime which adds memory and time overhead.

I think it might be better to just rip libuv out and put in a much more minimal replacement and then have this replacement be implemented using raw epoll and libuv.

The reason for having loops and poll callbacks in a global vector is because of similar reasons: libuv bloats its polls with 8 byte callbacks when most applications have less than 128 callbacks in total. uws has something like 16 callbacks so it makes no sense to store this in 8 bytes when you can store it using 4 bits

alexhultman:
Think about a middle layer that can be implemented jsing C++ standard networking, libuv or raw epoll

jimmyt857:
Ok, I have no problem with reorganizing this. I'm not quite sure what you mean in a few places though, could you clarify?

> I'm not a fan of prefixing everything with UWS_UV

The only alternative to this I can think of is to use something like unet::poll_t instead of UWS_UV uv_poll_t, and then we could set up two different definitions for unet based on the macros. Is that what you have in mind?

> I'm also not a fan of introducing handle as a runtime value to every uv_handle_t.

What does handle refer to?

> I think it might be better to just rip libuv out and put in a much more minimal replacement and then have this replacement be implemented using raw epoll and libuv.

My understanding was that this IS the much more minimal replacement. How would that differ from what I have here?

> Think about a middle layer that can be implemented jsing C++ standard networking, libuv or raw epoll

I don't think there is much room to split uUV into two separate layers. The lowest layer has to have the event loop, which must understand the concept of poll / timer / async handles. Could you give an example of what functionality would happen in this middle layer vs. the bottom layer?

alexhultman:
uv_handle_t has "type" as an actual runtime member which is absurd given that types should not be dynamic, they should be static. So it would be possible to get better memory usage by removing libuv and designing something better. The libuv interface has some design flaws that harm performance and memory usage.

uv_poll_t is 160 bytes big in libuv but only 32 bytes in uUV (the old version). By adding "type" as an actual integer attached to every uv_handle_t you add a very significant overhead to all polls, when the type is already known in the programming language (C++ is all about static-everything)

Static dispatch:
uv_close(uv_timer_t)
uv_close(uv_async_t)

Dynamic dispatch:
uv_close(uv_handle_t *) <- switch by "type" is dynamic and thus inferior to static "dispatch"

This is just one thing that is wrong with libuv

jimmyt857:
Ok, I've made these changes. I rearranged some of the members to optimize the packing, and uv_poll_t is still 32 bytes. What else can be improved?

alexhultman:
Definitely very nice and I want to merge it but I don't want these UWS_UV prefixes. My idea was you link to either libuv or uUV and then you won't get any conflicts. Maybe this is not optimal but for this merge to happen at this point I can't have 17 files changed -> it should be 2 or maybe 3-4. Later on we can think about how to solve this conflict thing.

jimmyt857:
But that method would prevent users from using libuv anywhere in their application if they use uUV, and we dont implement all of the libuv features. You don't want to retain support for this?

jimmyt857:
Just FYI, I just ran the benchmarks for the first time, and there seems to be a boost, particularly for memory. However, I'm running on an Amazon EC2 small instance, and I was not able to use more than 40k connections, so I would love to know what results you get. Here are my numbers:

Old version:
scalability 40000 3000
Connection performance: 23.7389 connections/ms
Memory performance: 891.675 connections/mb

throughput 10000 20 10 3000
Averages ~ 565-570 echoes / ms

New version:
scalability 40000 3000
Connection performance: 27.0636 connections/ms
Memory performance: 1205.13 connections/mb

throughput 10000 20 10 3000
Averages ~ 560-565 echoes / ms

Overall, small improvement to connection speed, large improvement to memory usage, and small hit to message throughput. I'm not really sure why the small message throughput went down for me. Again, I'm very curious to know what you get.

jimmyt857:
On second thought, there are 3 places I can think of to improve speed:

- Store callback pointers instead of indexes. This skips an array lookup. Of course, this would tradeoff an extra 8 bytes of memory / connection handle.
- Increase epoll_wait max events. The current value remains at 64, and I haven't played with it at all. Perhaps a larger value would allow for less overhead of the event loop.
- Avoid callback copying. Currently, the event loop copies all of the uv_timer_t* handles that are ready into a new vector before looping over them. This is required to that calls to uv_timer_start / stop within the callback to not invalidate the iterator. Instead, we could have an add / remove stage of the event loop, so that this copying would not be necessary. However, I'm not sure if this is an issue for the performance tests because this copying is not necessary for uv_poll_t*, due to epoll_wait giving us a separate array.

jimmyt857:
Using 1024 instead of 64 for max events, I get the following results:

scalability 40000 3000
Connection performance: 25.89 connections/ms
Memory performance: 1201.31 connections/mb

throughput 10000 20 10 3000
Averages ~ 570-575 echoes / ms

So overall it's another tradeoff, sacrificing a little bit of both connection speed and memory for message throughput.

jimmyt857:
Sorry for making so many posts, but I just noticed something strange.

When I change the number of connections in the scalability test, the memory performance value scales directly with it. For example, I get the following outputs:

scalability 100 3000
Connection performance: 12.5 connections/ms
Memory performance: 3.01318 connections/mb

scalability 1000 3000
Connection performance: 16.3934 connections/ms
Memory performance: 30.1318 connections/mb

After looking at the code, I think maybe the memory usage is being evaluated after all of the connections are closed, so the call to pmap simply gives a fixed value.

alexhultman:
You need to test with something like 1 mil connections to get accurate readings from that test. But it's quite easy to just add the sizeofs together. About 300 mb per 1 mil is about 300 bytes per connection which makes sense when uv_poll_t is 160 bytes, websocket state is 24 bytes, some other shit is the rest.

When it comes to UWS_UV -> yes the design currently is that you can only use libuv OR uUV. This is not optimal but I cannot merge 17 file changes in one PR -> if you make it in two steps we can think about how to solve the conflict thing. I want to have a middle layer that is somewhat like libuv but a little different

jimmyt857:
Ok, how about this: first commit, I take out all of the UWS_UV and comment out the uUV namespace. Everything will work, but as I said it will still have that problem. Second commit adds the namespace and either the prefixes or a middle layer.

alexhultman:
Yep maybe it is just me but I don't like it when a lot of files change all at once. If you can take out UWS_UV I can probably merge it straight in.

jimmyt857:
Haha alright, this shouldn't take long.

jimmyt857:
Ok, it's done. It still touches 8 files because of the slight interface change for uv_close.

alexhultman:
Oh, so you managed to remove "type" and use static dispatch in uv_close? Very nice

src/uUV.cpp
```diff
@@ -1,11 +1,356 @@
-#ifdef USE_MICRO_UV
-
 #include "uUV.h"
 
+#ifndef USE_MICRO_UV
+void uv_close(uv_async_t *handle, uv_close_cb cb) {
```
alexhultman: I don't get this? They all redirect it to the general uv_close?

jimmyt857: libuv uses dynamic dispatch but uUV uses static dispatch. The redirect is just to avoid having \#ifdef USE_MICRO_UV every time uv_close is called, because libuv requires the uv_handle_t * cast and uUV requires that we do not cast.

src/uUV.cpp
```diff
@@ -1,11 +1,356 @@
...
+        }
+    }
+    if (poll->cbIndex == pollCbHead) {
+        poll_callbacks[pollCbHead++] = cb;
+    }
+    return epoll_ctl(poll->get_loop()->efd, EPOLL_CTL_MOD, poll->fd, &poll->event);
+}
+
+int uv_poll_stop(uv_poll_t *poll) {
+    poll->flags &= ~UV_HANDLE_RUNNING;
```
alexhultman: Is this (and other similar things) really needed in µWS? We track this ourselves and there is no need to have uUV track it again. This applies to other places also, the idea is that you can strip down the implementation to the very minimal since we know how libuv is used and what features are used.

src/uUV.cpp
```diff
@@ -1,11 +1,356 @@
...
+    uv_poll_t *poll = (uv_poll_t *) handle;
+    poll->fd = -1;
+
+    handle->flags |= UV_HANDLE_CLOSING;
+    loop->closing.push_back({(uv_handle_t *) handle, cb});
+}
+
+int uv_fileno(uv_poll_t *handle) {
+    if (uv_is_closing(handle))
+        return UV_EBADF;
```
alexhultman: This is also overly correct. We never query the FD of a closing socket.

src/uUV.cpp
```diff
@@ -1,11 +1,356 @@
...
+    return epoll_ctl(poll->get_loop()->efd, EPOLL_CTL_DEL, poll->fd, &poll->event);
+}
+
+void uv_close(uv_poll_t *handle, uv_close_cb cb) {
+    uv_loop_t *loop = handle->get_loop();
+
+    uv_poll_t *poll = (uv_poll_t *) handle;
+    poll->fd = -1;
+
+    handle->flags |= UV_HANDLE_CLOSING;
```
alexhultman: This can be described as -1 fd, no need to hold a flag value for this

src/uUV.cpp
```diff
@@ -1,11 +1,356 @@
...
+
+    loop->asyncs.insert(async);
+}
+
+void uv_async_send(uv_async_t *async) {
+    uv_loop_t *loop = async->get_loop();
+    loop->async_mutex.lock();
+    uint64_t val = 1;
+    int w = write(loop->asyncWakeupFd, &val, sizeof(val));
+    async->run = true;
```
alexhultman: I also guess this is not strictly needed

alexhultman:
I added some comments. Basically the idea is to make uUV as minimal as it can possibly be, not as something 100% compatible with libuv (then it would just become a copy of libuv).

If we can starve uUV down to a minimal raw epoll wrapper and get µWS running on that then we have successfully removed a lot of bloat that we don't need. If we then wrap this minimal thing in a very strict interface we can implement this with the less strict libuv.

Consider uUV the Vulkan of graphics and libuv the OpenGL -> Vulkan is all about having the bare minimals and zero checks while OpenGL is more like a complete blob of checks and error handling and shit you don't need in production mode.

alexhultman:
One example: if you call uv_poll_stop on a poll that is already stopped in libuv, then it will not stop. In uUv it should just hard crash because this is up to the application to fix, not the low level library. It should just hard crash and burn if the application is broken -> just like Vulkan does :+1:

src/uUV.cpp
jimmyt857: The running flags are actually needed, for some specific edge cases of certain callbacks calling uv_stop/uv_close on other callbacks. The most obvious case right now is when a uv_timer_t, for example, calls uv_stop on a different uv_timer_t that is also ready to fire in the same loop iteration.

src/uUV.cpp
jimmyt857: Ok, will remove this

src/uUV.cpp
alexhultman: But if it is an edge case then we can just ignore that. It doesn't have to be identical in behavior. If a timer is cancelled at around the same time as it should fire then it makes sense to not bother being 100% strict about behavior since this is not what the application wants anyways. You know, rules are made to be broken.

src/uUV.cpp
jimmyt857: fd is unique to uv_poll_t, so the closing flag is needed for the other types. I could remove this bitwise or here, but then i would have to write a different uv_is_closing for each handle type. Is that what you want?

src/uUV.cpp
jimmyt857: run is needed to identify which async handles are ready to fire. The mutex and the eventfd wakeup event are shared for all of the asyncs in a loop, so there needs to be something to tell us which subset of asyncs to call.

alexhultman:
I think the best solution is to take a look at µWS the application and figure out what functionality it expects from libuv, not the other way around. These small annoyances that are due to how libuv works could be removed if we just rip out libuv and start on a new interface from the start.

Design goals: small polls, static dispatch, don't care about stuff that doesn't matter, don't car about sizes of stuff that doesn't matter: uv_loop_t, uv_async_t and other things that are local to the server. Only stuff that is local to the socket needs to be small.

Some kind of loop, some kind of poll, some kind of timer, some kind of async (low prio) <- that's pretty much all that is needed. This is the size priority:

uv_poll_t <- highest priority, needs to be as small it can theoretically be
uv_timer_t <- not as important, only lives for a short while and I plan to move all timers to the server
uv_async_t <- makes no difference at all
uv_loop_t <- makes no difference at all

src/uUV.cpp
jimmyt857: Actually, these edge cases are rare enough / shouldn't occur in  our lib that I think I'll just remove this stuff for now. It can easily be added later if there are issues.

jimmyt857:
Ok, I took out the UV_HANDLE_RUNNING flag for everything except the timers. Timers need it because they re-add themselves to the timer list after firing, and they need to know if uv_timer_stop was called during the callback.

alexhultman:
Do you have skype or something?

jimmyt857:
I do, my skype tag is jimmyt857 also.

alexhultman:
#471 

You know what? I need to release this v0.13 thing first of all and then I need to fix eventual bugs and then after that we can take a look at ripping out libuv and implementing three different backends: c++17, epoll and libuv. Version 0.14 could have this stuff implemented

jimmyt857:
Ok, should I set up the interface you described or just let this sit for awhile?

alexhultman:
It can sit, nice job updating the uUV support. There are other things we can do, for instance epoll_ctl is thread safe so we can also add a specific function changeThreadSafe(int event) and skip that whole async thing in Socket::write. It makes sense to remove libuv

jimmyt857:
Also, FWIW, I've been looking at how to integrate mTCP. It actually looks really easy, but there is an issue (https://github.com/eunyoung14/mtcp/issues/112) with eventfd not being supported. eventfd() is how we wake up the epoll_wait call in async_send, so it's a requirement.

jimmyt857:
Do you think it would be ok to pull this? Otherwise I'll have to go back and update things as master continues getting updated. Without USE_MICRO_UV there aren't any changes.

jimmyt857:
Awesome, thanks!

alexhultman:
uUV.cpp is missing implementation of uv_is_closing:

node: symbol lookup error: /home/alexhultman/uWebSockets/nodejs/dist/uws_linux_47.node: undefined symbol: _Z13uv_is_closingP9uv_poll_s

jimmyt857:
Hmmm, this should be it https://github.com/uWebSockets/uWebSockets/pull/465/files#diff-6fee0f6ce2eb80fcfb677bcb041203d8R204

jimmyt857:
oh no im dumb, hold on

alexhultman:
cf1991122f3f6b73f4fb98fb93ae86d67a85e5e4

PR #472, opened 2017-02-20
Allow listening on specific addresses

nischu7:
It should be possible to explicitly specify the address to bind to instead of just listening on all interfaces.

This change adds a "host" argument to `Hub::listen` which gets passed to `getaddrinfo()` in `Node.h` while preserving API compatibility.

### DIFF
--- a/src/Hub.h
+++ b/src/Hub.h
@@ -39,6 +39,7 @@ struct WIN32_EXPORT Hub : private uS::Node, public Group<SERVER>, public Group<C
     static void onClientConnection(uS::Socket s, bool error);
 
     bool listen(int port, uS::TLS::Context sslContext = nullptr, int options = 0, Group<SERVER> *eh = nullptr);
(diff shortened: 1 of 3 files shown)


src/Hub.h
```diff
@@ -39,6 +39,7 @@ struct WIN32_EXPORT Hub : private uS::Node, public Group<SERVER>, public Group<C
     static void onClientConnection(uS::Socket s, bool error);
 
     bool listen(int port, uS::TLS::Context sslContext = nullptr, int options = 0, Group<SERVER> *eh = nullptr);
+    bool listen(const char *host, int port, uS::TLS::Context sslContext = nullptr, int options = 0, Group<SERVER> *eh = nullptr);
```
alexhultman: If you swap port / host order you can set host = nullptr as default and remove the added function

alexhultman: Oh wait, I see. You preserved interface

alexhultman:
Good shit :+1:

PR #477, opened 2017-02-20
allocate loop structure on cache line boundary.

ronag:
Allocate loop struct on cache line boundary. This together with #476 will make the entire structure to fit and be aligned on 3 cache lines (3 x 64 bytes).

### DIFF
--- a/src/uUV.cpp
+++ b/src/uUV.cpp
@@ -50,7 +50,7 @@ uv_loop_t *uv_handle_t::get_loop() const {
 }
 
 inline uv_loop_t *uv_loop_helper() {
-    uv_loop_t *loop = new uv_loop_t;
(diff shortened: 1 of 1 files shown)


alexhultman:
Placement new on aligned_alloc does not play well with delete -> this introduces undefined behavior and memory leaks. Only uv_poll_t is important in this. Loops are of least importance (they could be allocated on a Viking Rune and still be okay) This is just basic experimentation code that will be transformed into the new interface first of all

alexhultman:
Like I said these PRs are not relevant at this point since all of uUV will be moved into a new interface where rules of what should be implemented will change. When this new interface has been established we can look at implementing this efficiently. It makes no sense to micro optimize something that will fundamentally change.

PR #484, opened 2017-02-21
Implement uUV api using value/move semantics.

ronag:
This requires more updates but so far I've only updated uUV to discuss the proposal. By using a value/move semantic based interface we should be able to both improve correctness as well as in the future possibly avoid head allocations in favor of copying small data structures.

### DIFF
--- a/src/uUV.h
+++ b/src/uUV.h
@@ -20,7 +20,7 @@ struct Loop : uv_loop_t {
         }
     }
 
-    void destroy() {
+    ~Loop() {
         if (this != uv_default_loop()) {
             uv_loop_delete(this);
         }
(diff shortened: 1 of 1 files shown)


alexhultman:
Stop posting these

ronag:
Fair enough. I'll stick with forking.

alexhultman:
Problem is you send a million PRs that you haven't tested, haven't understood the direction to take and in all or many cases change stuff that doesn't matter or simply just introduces more indirection. If you want things merged you need to follow the direction we are heading, test your changes and work on things that are prioritized

ronag:
@alexhultman: No worries! You are right. I'm just getting familiarized with the code and wanted to point out some ideas. I understand if you are busy and don't want to get spammed with random incomplete ideas from random people.

I'm going to play around more with this on my vacation and maybe come with some proper PR:s. Cheers!

Issue #489, opened 2017-02-22
Is there a memory leak on send errors?


GGRI:
Hi,

I have a question about uWS 0.13.0 (also 0.12.0).
On Socket.h, in **void sendTransformed** why don't you call **freeMessage** when **write** returns false?

It seems messagePtr is never freed and occurs a memory leak.  This is what I see when monitoring memory.

For me, `freeMessage(messagePtr);` should be added on line 449 in socket.h and `getSocketData()->nodeData->freeSmallMemoryBlock((char *)messagePtr, memoryIndex);` on line 429 in socket.h.

Sincerely,
Guillaume

alexhultman:
Hawk eye achievement :1st_place_medal:

GGRI:
Yeay :-)

There is the same behavior on WebSocket.cpp (0.13.0) on line 98.

0x111111111:
Awesome! Can't wait when 0.14 will be released :+1:

playablebox:
Is there anyway to know if this fix is in the nodejs version of uws 8.14.1?
I'm using artillery.io for load testing which creates 10 virtual users per second and each of them is configured to send 400 messages in 13s (1 msg every 0.033 seconds). 
A lot of memory is consumed (1GB in about 20min) and not released to OS (not even after killing the load test process).
If I just comment out wss.send, then it runs for hours without any leak.

Any ideas?

alexhultman:
@playablebox 

```
  config:
    target: "ws://localhost:3000"
    phases:
      - duration: 1800
        arrivalRate: 10
    ws:
      # Ignore SSL certificate errors
      # - useful in *development* with self-signed certs
      rejectUnauthorized: false
  scenarios:
    - engine: "ws"
      flow:
        - loop:
          - send: "hello"
          - think: 0.0325
          count: 400
```

I'm running this, from your description. So far there is no sign of any leak at all. It consumes 0% CPU time and exactly 476 kb of memory (no change whatsoever, completely flatlined). Artillery.io itself is choking at 100% CPU time and evergrowing memory usage (now up at 1 GB and growing).

This is the C++-only library, not the wrapper. So what you are observing is most probably the garbage collector being its usual, unreliable piece of shit. Just like artillery.io is at 1 GB it doesn't sound impossible uws would be too. Artillery.io opens and closes a lot of connections and that involves going through a lot of JavaScript code in Node.js which leaves a lot of work for the GC.

[...]

PR #500, opened 2017-02-27
Is there a memory leak on send errors? #489

lilitZak:
Fixed issue 
https://github.com/uWebSockets/uWebSockets/issues/489
[](https://github.com/uWebSockets/uWebSockets/issues/489)
Please check.


alexhultman:
Will take a look at this -> I know there are many places where this applies.

alexhultman:
This is not correct -> it causes double frees. I commited a fix separate from this

Issue #385, opened 2017-01-03
Unable to store handles to websockets when transferring connections


mafrost:
When using multithreading and transferring connections to other hubs using similar to following pseudo code, the `ws` instance seem to get deleted inside the transfer. This leads to it being impossible to track individual clients after transfer.

```
master.onConnection([this](uWS::WebSocket<uWS::SERVER> ws, uWS::UpgradeInfo ui) {
    auto wsd = new UserData(ui.path);
    ws.setUserData(reinterpret_cast<void*>(wsd));
    int i = randomIndex(); 
    this->myListOfWebsockets.push_back(ws);
    ws.transfer(&this->instances[i]->hub->getDefaultGroup<uWS::SERVER>());
}
```

some other place in the code

```
auto wsd = reinterpret_cast<WsUserData*>(this->myListOfWebsockets[0].getUserData());
// throws because the ws handle is incorrect
```

Do you have any idea how to solve this? Also maybe there should be some type of `onTransfer` event, similar to the `onConnection` event. Or perhaps the `onConnection` event should be called after a transfer inside that hubs event loop, that would lead to it being performed in the "correct" thread and make it easier to store the client handle.

mafrost:
Maybe modify the `WebSocket.h` file to:

```
    void transfer(Group<isServer> *group) {
        ((Group<isServer> *) getSocketData()->nodeData)->removeWebSocket(p);
        uS::Socket::transfer((uS::NodeData *) group, [](uv_poll_t *p) {
            uS::Socket s(p);
            ((Group<isServer> *) s.getSocketData()->nodeData)->addWebSocket(s);

            // start insert

            ((Group<isServer> *) s.getSocketData()->nodeData)->connectionHandler(s, {nullptr, nullptr, 0, 0});

            // end insert

        });
    }
```

this way, a onConnection event will be triggered with the "new/correct" websocket handle.

alexhultman:
Probably best to add a callback to WebSocket::transfer and add a void *userData to Socket::transfer in where to hold the cb.

mafrost:
Yeah that could work as well, however I think it makes more sense to have `transferHandler`, just like the `connectionHandler` and call that in the `transfer` function. This way, it's possible to have one definite handler for all transfers to a hub instead of having to add callbacks from all different hubs which may transfer the connection (almost the same but not really). I think that performance wise it would be better to have a `transferHandler`.

```
// same code as above but instead of connectionHandler
[...]

PR #503, opened 2017-03-01
add callback for transferred sockets

jgilje:
I was in need of a fix for #385 myself, and implemented this PR as simple solution


mafrost:
I made a similar solution in the end as well. It really is needed for multithreading to be interesting in a real world scenario.

alexhultman:
I personally prefer a callback more like the one in send (not tied to the group) but hey this is needed so let's add it

endel:
Is it possible to use this from NodeJS as well?

PR #505, opened 2017-03-01
Filter WebSocket connections before open

0x111111111:
Simple usage:

```c++
h.onHttpUpgrade([](uWS::HttpSocket<uWS::SERVER> s, uWS::HttpRequest req) {
    return false;
});
```

In this example WebSocket connections are not allowed.
If you provide no `httpUpgradeHandler` the connections are allowed by default.

Yep, `s.terminate()` would do the same. But consider this example:

```c++
h.onHttpUpgrade([](uWS::HttpSocket<uWS::SERVER> s, uWS::HttpRequest req) {
    return req.getHeader("origin").toString() == "http://example.com";
});
```

If the [`Origin` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) isn't `http://example.com`, connection is rejected.
Why this is better than checking it after the connetion is open? Because instead of adding it to the pool and removing it, we can close it before.

FYI, I didn't know how to name this feature properly, feel free to change it.

### DIFF
--- a/src/Group.cpp
+++ b/src/Group.cpp
@@ -233,7 +233,7 @@ void Group<isServer>::onCancelledHttpRequest(std::function<void (HttpResponse *)
 }
 
 template <bool isServer>
-void Group<isServer>::onHttpUpgrade(std::function<void(HttpSocket<isServer>, HttpRequest)> handler) {
+void Group<isServer>::onHttpUpgrade(std::function<bool(HttpSocket<isServer>, HttpRequest)> handler) {
     httpUpgradeHandler = handler;
 }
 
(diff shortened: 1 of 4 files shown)


alexhultman:
I need to rethink how the upgrade path should look like at some point. Maybe something like this will be added, maybe not. I don't know yet. Currently there are about 3 different paths and I would like to merge them all into one common path (Node.js addon from HTTP -> WS, Node.js addon from built-in HTTP -> WS, default upgrade path).

0x111111111:
This PR is on hold.

PR #509, opened 2017-03-03
Handle EWOULDBLOCK on Windows

jgilje:
Winsock don't set errno (ref. https://msdn.microsoft.com/en-us/library/windows/desktop/ms737828(v=vs.85).aspx), and defines the value under a different `#define` variable.

This PR handles both the unix-way and the windows-way to detect an EWOULDBLOCK condition.


alexhultman:
This is fixed now as a separate wouldBlock() function. Thanks, I must have removed some kind of `#define errno WSAGetLastError()` or it has always been broken, don't know which? Thanks for the PR.

Issue #399, opened 2017-01-20
Add MinGW support


ghost:
Is it possible to compile `uWS` using MinGW64 (downloaded using MSYS2) on Windows?
```bash
$ cmake .
-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is GNU 6.3.0
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
-- Check for working C compiler: /mingw64/bin/cc.exe
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
-- Check for working C compiler: /mingw64/bin/cc.exe -- works
-- Detecting C compiler ABI info
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
-- Detecting C compiler ABI info - done
-- Detecting C compile features
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
-- Detecting C compile features - done
-- Check for working CXX compiler: /mingw64/bin/CC.exe
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
-- Check for working CXX compiler: /mingw64/bin/CC.exe -- works
-- Detecting CXX compiler ABI info
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
System is unknown to cmake, create:
Platform/MINGW64_NT-6.2 to use this system, please send your config file to cmake@www.cmake.org so it can be added to cmake
System is unknown to cmake, create:
[...]

PR #510, opened 2017-03-03
mingw-w64 support

jgilje:
This is a PR for #399, and basicly just @ssj17vegeta's fixes in `Networking.h`. I have only tested against the libuv backend. I did not have any issues with crashes related to `Node.h`, but I'm running a branch which also integrates #509, which changes EWOULDBLOCK.

I've tested running both the HTTP server and WebSocket server parts of uWebSockets after applying this patch.


0x111111111:
Will it work with MinGW64?

jgilje:
Yeah, actually, this was with the mingw-w64 distribution (http://mingw-w64.org/doku.php) as shipped by the Qt project for their MinGW build. I believe the older MinGW distribution from http://mingw.org/ is deprecated now.

0x111111111:
Yep. But the title is "mingw**32** support". Please make sure it'll work with MinGW64 downloaded using MSYS2: http://www.msys2.org (make sure you're running the MinGW64 console, not the MSYS one).

jgilje:
@0x111111111 you're correct. I just assume -w64 these days as the original mingw project hasn't updated for a long time. I've updated the PR title to reflect this. I can confirm this was compiled and tested with mingw-console (and not MSYS).

The actual toolchain I tested with was the one provided by Qt 5.7, but as mentioned on http://wiki.qt.io/MinGW
> Newer Qt 4.8 binary packages ship with a mingw-w64 based toolchain. For Qt 5, a newer MinGW-w64 toolchain is actually required.

I can do an additional compile and test using MSYS2 with a mingw-console if you believe there's a difference.

0x111111111:
Additional compile is not needed. MinGW64 from MSYS can use Qt as well.

alexhultman:
This is not endian portable

jgilje:
@alexhultman good observation. However, Windows has always been tied to little-endian processors (and this block is guarded by _WIN32). Even the ports to various RISC processors booted in LE mode (they usually supported operating in both BE and LE mode).

However, if you're still not convinced, I'll add a BE/LE ifdef and revise the patch.

Btw, I also noticed GCC has a builtin called `__builtin_bswap64`, which could be used instead of the `bswap_` inlines in the patch.

alexhultman:
My mingw64 comes with __builtin_bswap64

alexhultman:
Ah right I though you wrote __builtin_bswap32. Yeah just toss in __builtin_bswap64 instead and just put some comment about endian

src/Networking.h
```diff
@@ -24,11 +24,27 @@
 #include <WinSock2.h>
 #include <Ws2tcpip.h>
 #define SHUT_WR SD_SEND
-#define htobe64(x) htonll(x)
-#define be64toh(x) ntohll(x)
 #define __thread __declspec(thread)
-#define pthread_t DWORD
-#define pthread_self GetCurrentThreadId
+#ifdef _MSC_VER
```
alexhultman: maybe remove this and just have it be the inverse of __MINGW32__

src/Networking.h
```diff
@@ -24,11 +24,27 @@
...
+      return (__x >> 8) | (__x << 8);
+    }
+
+    static __inline unsigned int bswap_32 (unsigned int __x) {
+      return (bswap_16 (__x & 0xffff) << 16) | (bswap_16 (__x >> 16));
+    }
+    static __inline unsigned long long bswap_64 (unsigned long long __x) {
+      return (((unsigned long long) bswap_32 (__x & 0xffffffffull)) << 32) | (bswap_32 (__x >> 32));
+    }
+    #define htobe64(x) bswap_64(x)
```
alexhultman: __builtin_bswap64 here then yeah

alexhultman:
This is how I would put it:

#ifdef __MINGW32__
// assumes host is LE
#define htobe64(x) __builtin_bswap64(x)
#define be64toh(x) __builtin_bswap64(x)
#else
#define htobe64(x) htonll(x)
#define be64toh(x) ntohll(x)
#define pthread_t DWORD
#define pthread_self GetCurrentThreadId
#endif

jgilje:
@alexhultman got it, my branch has been updated. I also moved the `#define __thread __declspec(thread)` away when compiling with mingw, as this build should use GCC's  `__thread`.

alexhultman:
:warning: This branch has conflicts that must be resolved

jgilje:
Sorry, resolved now. Forgot about it as it was just an image.

PR #522, opened 2017-03-13
Add libuWS.so and libuWS.dylib to .gitignore

AstralSorcerer:
Add a .gitignore file


alexhultman:
.gitignore was removed recently

AstralSorcerer:
Could you please explain why? It's a huge pain having git complain about untracked files.

Issue #524, opened 2017-03-13
[nodejs] Busy loop (100% CPU) with Node.js cluster after a connection error


feldgendler:
When using the `uws` module in a service that users Node.js `cluster`, a connection failure causes 100% CPU.

Here is a minimal test case:

```
'use strict';
const cluster = require('cluster');
const uws = require('uws');

if (cluster.isMaster)
{
    new uws('ws://localhost:1'); // will definitely fail
    cluster.fork();
}
```

This always reproduces the bug for me (Ubuntu 14.04.5 LTS, x86-64, Node.js 6.10.0). Note that `uws` is only invoked in the master process, and the child process does nothing at all. However, if `cluster.fork` is not called, the bug does not manifest.

When I run the above code, it starts using 100% CPU. The worker is idle (as expected). The master is busy doing this:

```
[pid 21529] epoll_ctl(5, EPOLL_CTL_DEL, 9, {EPOLLHUP, {u32=9, u64=9}}) = -1 EBADF (Bad file descriptor)
[pid 21529] epoll_wait(5, {{EPOLLHUP, {u32=9, u64=9}}}, 1024, -1) = 1
[pid 21529] epoll_ctl(5, EPOLL_CTL_DEL, 9, {EPOLLHUP, {u32=9, u64=9}}) = -1 EBADF (Bad file descriptor)
[pid 21529] epoll_wait(5, {{EPOLLHUP, {u32=9, u64=9}}}, 1024, -1) = 1
[pid 21529] epoll_ctl(5, EPOLL_CTL_DEL, 9, {EPOLLHUP, {u32=9, u64=9}}) = -1 EBADF (Bad file descriptor)
[pid 21529] epoll_wait(5, {{EPOLLHUP, {u32=9, u64=9}}}, 1024, -1) = 1
```

feldgendler:
Also, if `fork` is called before the connection attempt, the problem disappears.

This way, it works as expected:
```
'use strict';
const cluster = require('cluster');
const uws = require('uws');

if (cluster.isMaster)
{
    cluster.fork();
    new uws('ws://localhost:1'); // will definitely fail
}
```

`O_CLOEXEC` not set where it should?

feldgendler:
Adding `SOCK_CLOEXEC` when opening sockets seems to fix the problem. Pull request #525.

alexhultman:
#308 

[...]

PR #525, opened 2017-03-13
Set SOCK_CLOEXEC when opening sockets

feldgendler:
Fixes issue #524

### DIFF
--- a/src/Node.h
+++ b/src/Node.h
@@ -47,7 +47,7 @@ public:
             return nullptr;
         }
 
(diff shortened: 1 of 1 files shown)


alexhultman:
Linux and FreeBSD both have SOCK_CLOEXEC and Windows doesn't care. OS X doesn't have SOCK_CLOEXEC though, so there is a small gap where you can still leak FDs if you fork at the same time as you accept a new connection...

feldgendler:
But in Node.js you don't have other threads that can fork. So no race condition. Perhaps the code can even be simplified so that the OS X branch is used everywhere.

PR #531, opened 2017-03-15
Tinymem

alexhultman:


alexhultman:
Known issues: multithreading is broken, use released version until fixed

Issue #447, opened 2017-02-09
Allow passing an optional callback to server.close()


jondubois:
The Node.js servers (both TCP and HTTP) allow you to pass in a callback when you call `server.close(callback)` - The callback will get executed when the server is fully down (so the port is available again).

See https://nodejs.org/api/http.html#http_server_close_callback

This is mostly useful for doing automated tests (creating and destroying servers dynamically). Sometimes you want to wait until the server is fully shut down before launching another one in its place.

alexhultman:
Yep, should be added

kdex:
Could we please get the inverse, too? `ws` has s an optional callback in the constructor, for when the server is listening on the port, see [here](https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketserveroptions-callback).

PR #532, opened 2017-03-16
+ Deviation from ws @ `webSocketServer.close`

swhgoon:


alexhultman:
It is accurate but I plan to have this feature added shortly: #447

PR #538, opened 2017-03-21
nodejs: support server random port definition

mkozjak:
Both nodejs' http and ws libs support random server listening port setting.

In example, someone might like to do:
```js
var WebSocketServer = require('uws').Server;
var wss = new WebSocketServer({ port: 0 })
```

Without this patch, the server would just return silently, since the ```options.port``` check would fail erroneously.

Also, sane checks are added here.

Signed-off-by: mkozjak <kozjakm1@gmail.com>


mkozjak:
Something's wrong with history, I guess. Cancelling PR.

PR #539, opened 2017-03-22
nodejs: support random server port definition

mkozjak:
Both nodejs' http and ws libs support random server listening port setting.

In example, someone might like to do:
```js
var WebSocketServer = require('uws').Server;
var wss = new WebSocketServer({ port: 0 })
```

Without this patch, the server would just return silently, since the ```options.port``` check would fail erroneously.

Also, sane checks for server object creation are added here.

Signed-off-by: Mario Kozjak <kozjakm1@gmail.com>

### DIFF
--- a/nodejs/src/uws.js
+++ b/nodejs/src/uws.js
@@ -345,6 +345,12 @@ class Server extends EventEmitter {
     constructor(options, callback) {
         super();
 
+        if (!options)
+            throw new TypeError('missing options');
+
+        if (options.port == null && !options.server && !options.noServer)
+            throw new TypeError('invalid options');
+
         var nativeOptions = WebSocketClient.PERMESSAGE_DEFLATE;
         if (options.perMessageDeflate !== undefined) {
             if (options.perMessageDeflate === false) {
@@ -437,18 +443,16 @@ class Server extends EventEmitter {
             _upgradeReq = null;
         });
 
-        if (options.port) {
-            if (options.host) {
-                this.httpServer.listen(options.port, options.host, ()=> {
-                      this.emit('listening');
-                      callback && callback();
-                });
-            } else {
-                this.httpServer.listen(options.port,  ()=> {
-                      this.emit('listening');
-                      callback && callback();
-                });
-            }
+        if (options.host) {
+            this.httpServer.listen(options.port, options.host, ()=> {
+                  this.emit('listening');
+                  callback && callback();
+            });
+        } else {
+            this.httpServer.listen(options.port,  ()=> {
+                  this.emit('listening');
+                  callback && callback();
+            });
         }
     }
 
--- a/nodejs/src/uws.js
+++ b/nodejs/src/uws.js
@@ -348,7 +348,7 @@ class Server extends EventEmitter {
         if (!options)
             throw new TypeError('missing options');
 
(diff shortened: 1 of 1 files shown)


alexhultman:
I can't find any official documentation for this. It seems to be more of an implementation detail rather than something official. In any case, unless you have very strong arguments for supporting this I don't want to follow Node.js designs.

ronkorving:
Port 0 behavior is documented under https://nodejs.org/api/net.html#net_server_listen_port_hostname_backlog_callback

nodejs/src/uws.js
```diff
@@ -345,6 +345,12 @@ class Server extends EventEmitter {
     constructor(options, callback) {
         super();
 
+        if (!options)
+            throw new TypeError('missing options');
+
+        if (options.port == null && !options.server && !options.noServer)
```
alexhultman: is (undefined == null)? I would probably use === undefined

nodejs/src/uws.js
```diff
@@ -345,6 +345,12 @@ class Server extends EventEmitter {
     constructor(options, callback) {
         super();
 
+        if (!options)
```
alexhultman: braces missing here and below

nodejs/src/uws.js
```diff
@@ -437,18 +443,16 @@ class Server extends EventEmitter {
...
-                      callback && callback();
-                });
-            } else {
-                this.httpServer.listen(options.port,  ()=> {
-                      this.emit('listening');
-                      callback && callback();
-                });
-            }
+        if (options.host) {
+            this.httpServer.listen(options.port, options.host, ()=> {
```
alexhultman: you could also fix all previous style issues here (6 space indent, () => {}, etc). Prior PR messed it up

nodejs/src/uws.js
mkozjak: I agree!

nodejs/src/uws.js
mkozjak: Done. There's also other style issues like too long lines (some are more than 120 characters long). Don't know what's your style. I can fix that if you'd like me to.

nodejs/src/uws.js
```diff
@@ -432,33 +455,32 @@ class Server extends EventEmitter {
...
-                      callback && callback();
-                });
-            }
+        if (options.host) {
+            this.httpServer.listen(options.port, options.host, () => {
+                this.emit('listening');
+                callback && callback();
+            });
+        } else {
+            this.httpServer.listen(options.port,  () => {
```
alexhultman: If you pass server as option, no host and no port, then it will still listen. It shouldn't listen if you pass it a server. This line also still has double space ",spacespace() => {".

alexhultman: I think you should replace if (options.port) from the original with if (options.port !== undefined) instead of removing it completely.

mkozjak: You are right!

alexhultman:
:+1:

mkozjak:
Cheers!

@alexhultman: Could you also bump and publish nodejs package version to 0.15.0 if possible?

alexhultman:
0.14.0 is the next version and it will release probably this week some time. You barely made the window -> releases take months of work.

PR #549, opened 2017-03-27
Issue 1: Add a cmake build system.

serialprimate:
Hi Alex,

I am aware that uWS did have a Cmake build of some description until recently, but I don't know much about it. I do know that I would like to start using uWS because it looks pretty good on paper.

I have created a new Cmake build in a fork of your repository. It includes an option to build with libuv as that seems to be needed for some systems. In my case (similar to that of many others!), I need to integrate with other libraries and have some non-standard locations of third party dependencies. At the moment, a short and sweet Cmake setup is the way to go for me to easily configure a build on the command line. I am hoping that you like it so I am sharing it. So far it is small, neat and does the job well:
```
$ mkdir build
$ mkdir install
$ cd build
$ cmake -Wdev --warn-unitialized -D CMAKE_BUILD_TYPE=RELEASE -D CMAKE_INSTALL_PREFIX=../install -DWITH_LIBUV=FALSE -G "Unix Makefiles" ..
$ make install
```
Thanks for taking the time to have a look.

Matt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,68 @@
+# CMake build configuration.
+
+# Generate makefiles with:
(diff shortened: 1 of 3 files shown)


alexhultman:
CMake is banned for all eternity here

sevenbitbyte:
Seriously guys? I'm trying to integrate uWebsockets right now into a CMake only eco-system. Some basic CMake support from the library maintainers would be much appreciated.

Finally it would be very helpful new comers if the `examples` included example builds for make and cmake. Finally I'm not sure why for testing purposes Qt's build system is prefered over simple makefile or cmake?

PR #555, opened 2017-03-29
Timer bug

jimmyt857:
I was getting segfaults when using Timers in my own code, with very low repeat values. If an event loop iteration takes longer than a repeat value, the timer will have a timepoint in the past when it is re-added. This happens because loop->timepoint is not updated. So, I have fixed this by updating loop->timepoint in the Timer::start call.

While I was already here, I also put in the insertion sort.

### DIFF
--- a/src/Epoll.h
+++ b/src/Epoll.h
@@ -72,16 +72,16 @@ struct Timer {
     }
 
     void start(void (*cb)(Timer *), int timeout, int repeat) {
-		loop->timepoint = std::chrono::system_clock::now();
+        loop->timepoint = std::chrono::system_clock::now();
         std::chrono::system_clock::time_point timepoint = loop->timepoint + std::chrono::milliseconds(timeout);
 
-		Timepoint t = {cb, this, timepoint, repeat};
-		loop->timers.insert(
-			std::upper_bound(loop->timers.begin(), loop->timers.end(), t, [](const Timepoint &a, const Timepoint &b) {
-				return a.timepoint < b.timepoint;
-			}),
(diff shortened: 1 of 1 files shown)


jimmyt857:
Just to clarify, the segfault occurs because of an infinite loop eventually resulting in memory errors. If a timer is re-added with a timepoint in the past, it will be immediately called again by Loop::run's timer loop.

alexhultman:
Nice. This can be improved in terms of performance (libuv guys claim getting the timepoint is heavy weight so they don't update it in start) but since we only use timers in some cases it shouldn't matter that much. The timer thing is real messy and I recently saw they have this timerfd thing in the Linux kernel that actually implement timers completely in the kernel so maybe I will look at that but yeah most important right now is to fix any segfaults.

alexhultman:
Hmmmmmmmm.. now that start is getting the timepoint every time it might actually harm performance significantly if many WebSockets are shutting down at the same time. You cannot look to improve this to only get the timepoint once and pass it along or something?

jimmyt857:
Well, if you only get the timepoint once then you can still get the same segfault, if the timer callback is a longer operator than your repeat. The only completely safe alternative I can think of is to re-add the timers after the loop, instead of inside it. So we would add them to a temporary list and then call start on each of them.

alexhultman:
I did a quck benchmark of std::chrono::system_clock::now() and it doesn't seem to have any significant overhead whatsoever. Makes sense not to listen to libuv people since they can't tell their mouth from their asshole. Nice fix.

PR #556, opened 2017-03-30
clear pollQueue onEnd

jgilje:
I'm using the libuv backend, and get an assertion fail from libuv. I'm testing my server for handling aborted transfers. To test this, I telnet onto the web-server, request a file and exit the telnet session. The server has a breakpoint set early during processing of the HTTP request, and resumes execution after I've closed the telnet session.

The server will call `changePoll(this);` in Socket.h on line 343, which later causes libuv to fail with the following assertion:
libuv/src/unix/poll.c:105: uv_poll_start: Assertion `!uv__is_closing(handle)' failed.

### DIFF
--- a/src/HTTPSocket.cpp
+++ b/src/HTTPSocket.cpp
@@ -296,6 +296,8 @@ void HttpSocket<isServer>::onEnd(uS::Socket *s) {
         delete httpSocket->preAllocatedResponse;
     }
 
+    httpSocket->nodeData->changePollQueue.clear();
+
     if (!isServer) {
         httpSocket->cancelTimeout();
         Group<CLIENT>::from(httpSocket)->errorHandler(httpSocket->httpUser);
--- a/src/HTTPSocket.cpp
+++ b/src/HTTPSocket.cpp
@@ -297,13 +297,10 @@ void HttpSocket<isServer>::onEnd(uS::Socket *s) {
     }
 
     httpSocket->nodeData->asyncMutex->lock();
-    auto end = httpSocket->nodeData->changePollQueue.end();
-    auto it = std::find_if(httpSocket->nodeData->changePollQueue.begin(), end, [=](Poll* s) {
-        return (s == httpSocket);
-    });
-    if (it != end) {
-        httpSocket->nodeData->changePollQueue.erase(it);
-    }
+    httpSocket->nodeData->changePollQueue.erase(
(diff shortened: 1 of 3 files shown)


alexhultman:
Nice find but you can't clear all, that vector is per group. And you also need to lock the asyncMutex to clear it (I think). This only happens if you send on a socket from a different thread, maybe better to remove that support in libuv as it complicates things

jgilje:
Yeah, I see. I thought nodeData was unique per HttpSocket. How about doing `setShuttingDown(true);` in `HttpSocket<isServer>::onEnd`, and checking this flag when iterating changePollQueue (And for completeness, also while iterating transferQueue)?

I've already implemented and tested this solution, and it works, but I'm thinking better check with you first this time :)

I don't believe I'm familiar enough with the codebase to implement your suggestion of removing the libuv support for multiple threads.

alexhultman:
No because after onEnd the socket is not guaranteed to exist anymore (it gets deleted next iteration of its loop). You need to remove all references to this socket in the changePollQueue but leave those entries not being this socket

jgilje:
Ok, just ammended the PR based on your input. Now, I assume at most one instance of the ending httpSocket can be found in changePollQueue. Let me know if this assumption is wrong.

alexhultman:
This whole changePoll thing needs work on libuv. Currently you can have multiple entires in the changePollQueue of the same Poll (makes no sense). So you will need to loop over all entries and remove all entires that are you:

for (auto it = changePollQueue.begin(); it != changePollQueue.end(); ) {
if (it->poll == this) {
it = changePollQueue.erase(it);
} else {
it++;
}
}

jgilje:
Alright! Updated to remove all instances of the socket which is ending

alexhultman:
Would you mind putting that in a function something like NodeData::clearPendingPollChanges(Poll *) and also add that call to WebSocket::onEnd (same flaw applies for websockets). Also, can you explain more about why you hit this bug? changePoll only queues things if it is called from another thread than the one it belongs to (as you can see in the code of changePoll it compared thread ids). You are using multithreading, correct? Does it seem to work otherwise?

jgilje:
Just pushed an updated solution with your feedback.

Well, changePoll() is called from Socket.h line 343, after I've disconnected the client. changePoll checks with threadSafeChange(). I'm using the libuv backend in my solution (I prefer developing on Linux, but my solution is required to also run on Windows, so I prefer having the same backend for both platforms). In Libuv.h, threadSafeChange() unconditionally returns false, and the request is handled by a different thread than the thread which new'd uWS::Hub - and so changePollQueue is modified.

PR #581, opened 2017-04-17
Check error properly

mingc00:
This pull request contain following modifications:
* Handling ::connect error
* Checking return value from recv function properly

There are some issues reported by static analysis tool. But I am not sure that they should be fixed or not.
1. Error from setsockopt is not handled.
2. NodeData and Socket have derived classes, but they don't contain a virtual destructor.

### DIFF
--- a/src/Node.h
+++ b/src/Node.h
@@ -103,7 +103,7 @@ public:
             return nullptr;
         }
 
-        if (::connect(fd, result->ai_addr, result->ai_addrlen) == -1) {
+        if (::connect(fd, result->ai_addr, result->ai_addrlen)) {
             freeaddrinfo(result);
             return nullptr;
         }
--- a/src/Node.h
+++ b/src/Node.h
@@ -103,7 +103,11 @@ public:
             return nullptr;
         }
 
-        ::connect(fd, result->ai_addr, result->ai_addrlen);
+        if (::connect(fd, result->ai_addr, result->ai_addrlen) == -1) {
+            freeaddrinfo(result);
+            return nullptr;
+        }
+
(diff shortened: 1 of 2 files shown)


src/Node.h
```diff
@@ -103,7 +103,11 @@ class WIN32_EXPORT Node {
             return nullptr;
         }
 
-        ::connect(fd, result->ai_addr, result->ai_addrlen);
+        if (::connect(fd, result->ai_addr, result->ai_addrlen) == -1) {
```
alexhultman: -1 should (pedantically) be SOCKET_ERROR (even though it is really just a macro for -1) on Windows. Or != 0.

src/Socket.h
```diff
@@ -260,7 +260,7 @@ struct WIN32_EXPORT Socket : Poll {
             int length = recv(socket->getFd(), nodeData->recvBuffer, nodeData->recvLength, 0);
             if (length > 0) {
                 STATE::onData((Socket *) p, nodeData->recvBuffer, length);
-            } else if (length <= 0 || (length == SOCKET_ERROR && !netContext->wouldBlock())) {
+            } else if (length == 0 || (length == SOCKET_ERROR && !netContext->wouldBlock())) {
```
alexhultman: If you bother, it could be !length instead of == 0 but that's maybe just me

src/Node.h
alexhultman: Yeah, probably best to just write if (connect) since that is less platform specific

mingc00:
Thanks for your reviews. I committed the updates.

PR #641, opened 2017-06-27
Simplify MSVC builds.

JunielKatarn:
What this change does:

1. Include OpenSSL and ZLib as NuGet dependencies (fetches headers and static libraries).
2. Add a Visual Studio solution file (uWebSockets), required for NuGet operations.

Note, libuv still needs to be resolved manually.

The Windows build becomes simpler, by just running two commands:
```
nuget restore
msbuild
```


saper:
Not sure if adding 80 lines counts as "simplify" :) We could NuGet things anyway without lots of XML I guess

alexhultman:
Vcpkg is what I push for. Nuget is some kind of Microsoft wannabe misunderstanding of what a package manager is.

PR #652, opened 2017-07-07
fix a bug about multiple errors emitted for one connection 

zhanglix:
fix a bug about multiple errors emitted for one connection which make resource management of userDate difficult.


src/Hub.cpp
```diff
@@ -114,6 +114,7 @@ void Hub::connect(std::string uri, void *user, std::map<std::string, std::string
             secure = true;
         } else if (protocol != "ws") {
             eh->errorHandler(user);
+            return;
```
alexhultman: what happens if this is not here? it gives two errors?

alexhultman:
Thanks

I still don't understand how I could edit your branch like that but it worked

PR #671, opened 2017-08-10
Add hostname parsing for IPv6 addresses to Hub::connect

jessetane:


alexhultman:
👍 neat

Sunstradamus:
There's a bug in this PR

src/Hub.cpp:95:28: error: ISO C++ forbids comparison between pointer and integer [-fpermissive]

That line should be

if (strcmp(uri[offset],"[") == 0) {

jessetane:
Thanks for the review, I will send another patch.

jessetane:
@Sunstradamus any idea why my compiler doesn't show the error you posted? I tried your code but it didn't compile.

Sunstradamus:
I'm using gcc 4.8.4 to compile. If you're going to use strcmp, you'll need to import cstring.

A better way would be to use uri.compare(offset, string::npos, "["). I wasn't sure if uri was a cstring or std::string

jessetane:
See #672 for possible fix

PR #675, opened 2017-08-10
Add hostname parsing for IPv6 addresses to Hub::connect

jessetane:


alexhultman:
It is not correct. What you want is something like find_first_of("]:/") regardless of ipv6 or ipv4. However if you got [ then expect a closing ] error otherwise. Current logic is not correct, it does not catch invalid URIs like this one:

ws://[yolo:3000/

alexhultman:
if current char is [ then ipv6 = true, step 1 char right.

find ] or : or /

if ipv6 then expect ] next, if so step 1
if : then read port, step,
if / then read path, done
if anything else, error

alexhultman:
Never mind that, this is not even correct either

jessetane:
Is it our job to catch every possible error the user could make entering the url? The call to `uS::Node::connect` will trigger the error handler if they get it wrong, and since it crashes the way it is now, I figure this is better than nothing.

jessetane:
It looks like the cause of the original crash is stoi bailing on a non-numeric port. This could happen IPv6 or no. I'll add a test for the URI you mentioned here and open a separate  PR to discuss the non-numeric port issue.

jessetane:
Check out #676

jessetane:
fixed by e402f022a43d283c4fb7c809332951c3108d6906

PR #676, opened 2017-08-10
Fix non numeric port crash

jessetane:


jessetane:
fixed by e402f022a43d283c4fb7c809332951c3108d6906

PR #701, opened 2017-09-10
HAVE_REUSEPORT: Other TCP/IP stacks may have SO_REUSEPORT too

saper:
Do not use __linux for this feature.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ CPP_OSX := -stdlib=libc++ -mmacosx-version-min=10.7 -undefined dynamic_lookup $(
 default:
 	make `(uname -s)`
 Linux:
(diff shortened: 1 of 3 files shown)


saper:
All tests passed fine on Linux with this code.

alexhultman:
You are moving the same OS check from source code to makefile

saper:
Yes, because with #705 I am adding FreeBSD and it makes little sense to add `|| defined (__FreeBSD__) || defined (somethingelese)` every time in the source code. Actually I think OS X has it available too (cannot test here).

alexhultman:
So make it a macro - the build script should not tell the source what to do, it should simply build the sources. From what I have gathered SO_REUSEPORT on freebsd os x is not the same as on Linux

PR #704, opened 2017-09-10
Use /bin/sh compatible redirect to quell stderr

saper:
If not using bash all testHTTP() requests will fail.


otamachan:
Hi, this change seems to redirect only ``stderr`` to ``/dev/null``. I think it will be helpful for debugging, though.
Thanks.

saper:
@alexhultman what's the problem with this pull request?

PR #711, opened 2017-09-13
Avoid re-defining SOCKET_ERROR and INVALID_SOCKET if they're already defined

sepehr-laal:
Macros `SOCKET_ERROR` and `INVALID_SOCKET` are commonly defined by other packages (e.g. look at WebRTC's source: [webrtc/rtc_base/socket.h](https://chromium.googlesource.com/external/webrtc/+/master/webrtc/rtc_base/socket.h#116)). Avoid redefining if it exists. If it exists and the value is not what we expect, throw a compile-time error.

### DIFF
--- a/src/Networking.h
+++ b/src/Networking.h
@@ -61,8 +61,23 @@ inline SOCKET dup(SOCKET socket) {
 #include <unistd.h>
 #include <arpa/inet.h>
 #include <cstring>
-#define SOCKET_ERROR -1
-#define INVALID_SOCKET -1
+
+#ifndef SOCKET_ERROR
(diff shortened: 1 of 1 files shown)


alexhultman:
You could also just prefix them with UWS_

sepehr-laal:
I don't have an issue with that, just did not want to touch the rest of the code. If that's what you prefer, I can update the PR.

alexhultman:
Yes better to solve a problem than to add a fancy error message. You can refactor these things in a good IDE with 2 mouse clicks. Static code = if it compiles, you're done

PR #731, opened 2017-10-23
UV_READABLE and UV_WRITABLE -> UWS_READABLE, UWS_WRITABLE

Sunstradamus:
Fixes name collision when compiling with libuv headers

### DIFF
--- a/src/Asio.h
+++ b/src/Asio.h
@@ -4,8 +4,8 @@
 #include <boost/asio.hpp>
 
 typedef boost::asio::ip::tcp::socket::native_type uv_os_sock_t;
-static const int UV_READABLE = 1;
-static const int UV_WRITABLE = 2;
+static const int UWS_READABLE = 1;
+static const int UWS_WRITABLE = 2;
 
 namespace uS {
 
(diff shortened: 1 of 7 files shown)


desertkun:
Thank you for this!

alexhultman:
So the issue is when you combine libuv with for instance Epoll?

alexhultman:
Are you having libuv loops and asio/epoll loops in the same program?

desertkun:
No, it was simply `libuv`. On linux in my case. So I guess Epoll is being included wrong.

Also, you have several issues with CMakeLists as `add_library` has old/missing files included.

alexhultman:
You are compiling with CMake? Who gave you CMake files?

desertkun:
Yeah, I am using CMake with `CMakeLists.txt` from your repository.

Here's my fixed lines:
```
add_library(uWS src/Extensions.cpp src/Group.cpp src/Networking.cpp src/Hub.cpp src/Node.cpp src/WebSocket.cpp src/HTTPSocket.cpp src/Socket.cpp src/Epoll.cpp)
```

alexhultman:
This repo has no CMake files and hasn't had for probably 1 year now. This sounds more like some kind of major mix up rather than an actual source problem.

desertkun:
Yeah, sorry for misunderstanding. Seems like that was solely my problem.

alexhultman:
I have reset the merge.

Sunstradamus:
> So the issue is when you combine libuv with for instance Epoll?

Yes whenever you use libuv with Epoll or ASIO. Specifically [this line](https://github.com/libuv/libuv/blob/695afe8322e4e9d8fbd8fe3f30069bca31b74f11/include/uv.h#L723) causes a duplicate identifier when compiling.

```c
#include <uv.h>
#include <uWS/uWS.h>

int main() {
    uWS::Hub h;
    return 0;
}
```

Sunstradamus:
I was originally using `USE_LIBUV` but after some random segfaults and freezing I poked around and realized that it wasn't thread safe unless compiled with Epoll + `UWS_THREADSAFE`

PR #734, opened 2017-10-25
Add onTransfer for clients on Hub class

ramonberrutti:
Easy way to make a onTransfer from hub.

If not you need to make something like this:
```
        m_Hub->getDefaultGroup<uWS::CLIENT>().onTransfer([this]( uWS::WebSocket<uWS::CLIENT> *ws ) {
            std::cout << "WebSocket: Client Transfer" << std::endl;
            m_ws = ws;
        });
```

instance of:
```
	m_Hub->onTransfer([this](uWS::WebSocket<uWS::CLIENT> *ws) {
            std::cout << "WebSocket: Client Transfer" << std::endl;
            m_ws = ws;
        });
```

### DIFF
--- a/src/Hub.h
+++ b/src/Hub.h
@@ -71,6 +71,7 @@ public:
     using Group<SERVER>::onConnection;
     using Group<CLIENT>::onConnection;
     using Group<SERVER>::onTransfer;
+	using Group<CLIENT>::onTransfer;
     using Group<SERVER>::onMessage;
     using Group<CLIENT>::onMessage;
     using Group<SERVER>::onDisconnection;
--- a/src/Hub.h
+++ b/src/Hub.h
@@ -71,7 +71,7 @@ public:
     using Group<SERVER>::onConnection;
     using Group<CLIENT>::onConnection;
     using Group<SERVER>::onTransfer;
-	using Group<CLIENT>::onTransfer;
+    using Group<CLIENT>::onTransfer;
     using Group<SERVER>::onMessage;
     using Group<CLIENT>::onMessage;
     using Group<SERVER>::onDisconnection;


alexhultman:
How good do you think this went?

ramonberrutti:
Hi @alexhultman I use in a game server (CSGO) that I need to sync the game loop with the ws and work perfect with clients. On server ws side, I also do many test for load balance and also work perfect.

```
cell_t ConnectWebSocket(IPluginContext *pContext, const cell_t *params) {
	char *text;
    pContext->LocalToString(params[1], &text);
    
    std::cout << "Connecting to: " << text << std::endl;

    uWS::Hub h;

    h.getDefaultGroup<uWS::CLIENT>().listen(uWS::TRANSFERS);

    h.onConnection([&h](uWS::WebSocket<uWS::CLIENT> *ws, uWS::HttpRequest req) {
        ws->transfer( &g_EBaconExtension.GetHubWs()->getDefaultGroup<uWS::CLIENT>() );

        h.getDefaultGroup<uWS::CLIENT>().close();
    });

    h.onDisconnection([&h](uWS::WebSocket<uWS::CLIENT> *ws, int code, char *message, size_t length) {
        h.getDefaultGroup<uWS::CLIENT>().close();
    });

    h.connect(text, (void *)10 );
    h.run();

    return 1;
}
```

Yes, that connection need to be in another thread.

mikeandtherest:
Hi there, and sorry for asking this here, but apparently I can't find any info on it: what is the difference and maybe advantage of using onTransfer() + transfer() over onMessage() + send()? Since I'm familiar with Socket.io, I've been using so far the latter, but I'm curious about the former, so some brief description of it would be really helpful. Thanks!

ramonberrutti:
@mikeandtherest transfer and onmessage don't is the same. message is the content and all the data that you receive. transfer you will use to transfer the ws to different hubs... Example you want loadbalance, o the use that i'm give for move the socket to another thread.

mikeandtherest:
@ramonberrutti Ah, ok. So transfer() is used not to send data, but to pass an actual ws to a group belonging to another hub, from a different thread. I've also read the 'testTransfers()' function from inside the tests file and I understand it now, but I wonder how does this help? I mean after all uWS handles requests in an async way by default, right? So then why would I want or need to load balance it using multithreading, i.e. by passing connections to the hub of another thread? I'm sure there's a good reason for this, just that I can't wrap my head around it. Thanks a lot for you time and help!

alexhultman:
Quoting the source comments:

/*
 * Transfers this WebSocket from its current Group to specified Group.
 *
 * Receiving Group has to have called listen(uWS::TRANSFERS) prior.
 *
 * Hints: Useful to implement subprotocols on the same thread and Loop
 * or to transfer WebSockets between threads at any point (dynamic load balancing).
 *
 * Warning: From the point of call to the point of onTransfer, this WebSocket
 * is invalid and cannot be used. What you put in is not guaranteed to be what you
 * get in onTransfer, the only guaranteed consistency is passed userData is the userData
 * of given WebSocket in onTransfer. Use setUserData and getUserData to identify the WebSocket.
 */

mikeandtherest:
> or to transfer WebSockets between threads at any point (dynamic load balancing)

This is what actually confuses me. How exactly does it help to have other hubs running on other threads? As far as I know uWS is async, so a client request doesn't prevent another client request to be received and handled. They can all be served in the same time, and that's what the event loop is for, right?
Thanks for your reply!

alexhultman:
An event loop works like this:

while (true) {
// block thread until event (0% CPU time until event)
epoll_wait(...);
// loop over ready sockets one by one, serially (100% CPU time until done)
for (...) {
// parse, emit & "handle" event
}
}

The more threads you have, the more CPU cores you can use to loop over the available sockets and parse their messages and run all that application logic.

mikeandtherest:
Gotcha! It makes a lot of sense now. So basically I'm making the server app running on multiple threads and handle events/requests in a distributed way. I have just two more question starting from this:

1) Even if I have multiple threads holding different hubs, I still have to have the main hub on the main thread, which acts like a "router" (or load balance manager) and which transfers all the connections to the other threads. But for a huge number of connections, can that main hub become a bottleneck at some point, since all the new connections are passing through him?

2) Is there a recommended number of threads to start with for a server using uWS? Like for a chat or collaborative whiteboard app? Or is it better to start with just one and dynamically add more (threads+hubs) based on some loading thresholds?

I really appreciate all your helps here, so thanks a lot again!

alexhultman:
There is an example: https://github.com/uNetworking/uWebSockets/blob/master/examples/multithreaded_echo.cpp

But no, there is no need to do threading in most cases.

mikeandtherest:
Hmm, that's interesting. So in Linux I can start multiple threads holding hubs listening on the same (reused) port. That's neat, by using that I don't have to transfer a ws from one hub to another. Thanks!

PR #740, opened 2017-11-16
Fixed -Wall -Wextra warnings

ksergey:
Fixed `-Wall -Wextra` compiling warnings

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-CPP_SHARED := -std=c++11 -O3 -I src -shared -fPIC src/Extensions.cpp src/Group.cpp src/Networking.cpp src/Hub.cpp src/Node.cpp src/WebSocket.cpp src/HTTPSocket.cpp src/Socket.cpp src/Epoll.cpp
+CPP_SHARED := -Wall -Wextra -std=c++11 -O3 -I src -shared -fPIC src/Extensions.cpp src/Group.cpp src/Networking.cpp src/Hub.cpp src/Node.cpp src/WebSocket.cpp src/HTTPSocket.cpp src/Socket.cpp src/Epoll.cpp
 CPP_OPENSSL_OSX := -L/usr/local/opt/openssl/lib -I/usr/local/opt/openssl/include
 CPP_OSX := -stdlib=libc++ -mmacosx-version-min=10.7 -undefined dynamic_lookup $(CPP_OPENSSL_OSX)
 
--- a/src/Epoll.h
+++ b/src/Epoll.h
@@ -1,6 +1,7 @@
 #ifndef EPOLL_H
 #define EPOLL_H
 
+#include "Utils.h"
 #include <sys/epoll.h>
 #include <sys/eventfd.h>
 #include <unistd.h>
@@ -45,6 +46,7 @@ struct Loop {
     void *preCbData, *postCbData;
 
     Loop(bool defaultLoop) {
+        ignore_unused(defaultLoop);
         epfd = epoll_create1(EPOLL_CLOEXEC);
         timepoint = std::chrono::system_clock::now();
     }
--- a/src/HTTPSocket.h
+++ b/src/HTTPSocket.h
@@ -1,6 +1,7 @@
 #ifndef HTTPSOCKET_UWS_H
 #define HTTPSOCKET_UWS_H
 
+#include "Utils.h"
 #include "Socket.h"
 #include <string>
 // #include <experimental/string_view>
@@ -173,10 +174,12 @@ struct HttpResponse {
 
         struct NoopTransformer {
             static size_t estimate(const char *data, size_t length) {
+                uS::ignore_unused(data);
                 return length;
             }
 
             static size_t transform(const char *src, char *dst, size_t length, int transformData) {
+                uS::ignore_unused(transformData);
                 memcpy(dst, src, length);
                 return length;
             }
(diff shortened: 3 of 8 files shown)


alexhultman:
This PR doesn't solve anything. Why are you enabling all warnings only to then silence them? A silenced warning is no better than an ignored warning. I'd rather have the compiler tell me these things as extra information *when I want to know* vs. permanently silence them just because your teacher said so.

ksergey:
A silenced warnings is better than ignored warnings because in some cases it's bug in code.
It's bad practice to leave warnings in code (even with -Wextra -Wall).

alexhultman:
Agree to disagree.

jimmyt857:
Just my 2 cents- silencing warnings is a bad idea, but some people (not me, but some people  =D) who compile their own libraries with -Wall -Werror will have to change their own compilations to avoid errors in uWS, which is not a good idea.

However, OP's method for fixing this is really poor design. All you need to do is comment out the name of the variables in the function header (i.e. ```Loop(bool /*defaultLoop*/) {```).

dvergeylen:
Just my 2 cents: silencing warnings is a bad idea, but declaring anonymous arguments like `f(int)` is valid. It simply tells the compiler "I know there are arguments here but I don't use them for now" and it won't start yelling.

I indeed compile my own programs with `-Wall -Wextra -Werror` and would benefit from such clarification in the source code.

@alexhultman, I can provide a pull request without `-Wall -Wextra` flags in the Makefile but commenting arguments as @jimmyt857 suggests if you would agree to merge it (tell me).

alexhultman:
The names of parameters are there to give better auto-completion / "intellisense" and to know what to put in. Removing the names will definitely be a negative thing with only a tiny bit of improvement.

alexhultman:
Btw I can get the code to error on compilation with -Wunsafe-loop-optimizations -Werror too. And that would not be solved by this PR. You can always find edge cases where some users will be unsatisfied. I have learnt it is a slippery slope to give in to every single edge case because it only means you will be spending all your time being pedantic for others.

alexhultman:
Basically the problem is -Werror, whenever you treat warnings as errors you probably need to chill out a bit and have some wine and chocolate. Warnings are not errors, they are warnings.

dvergeylen:
Haha, you are right but not everyone is as good C++ coder as you!

Another way not to worry about the unused variables is to include uWebsockets via `-isystem` instead of `-I` like this : `-isystem path/to/uWebSockets/src`. It will consider the headers as system headers and won't rise warnings as it consider you have no control on them ([gcc doc](https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html)).

From `man g++`:
>-isystem dir
>  Search dir for header files, after all directories specified by -I but before the standard system directories.  Mark it as a system directory, so that it gets the same special treatment as is applied to the standard system directories.

ksergey:
@alexhultman Warnings is warn that is possible bugs in code. False-positive warnings should be fixed, because they could hide real warnings in compiler output.

alexhultman:
@ksergey So are you willing to fix all warnings with -Wunsafe-loop-optimizations too? Because otherwise I would like to use the hypocrisy card. And then also go through all different compiler flags and fix all warnings that show up?

ksergey:
Only `-Wall -Wextra`

alexhultman:
@ksergey So why should we care for **your** use of `-Wall -Wextra` but ignore the very same "all warnings are bad and needs to be fixed immediately" logic for users who want to put in, say, `-Wunsafe-loop-optimizations`?

alexhultman:
This is hypocrisy and narcissism.

ksergey:
Ok `-Wall` enought. 

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options

```
This enables all the warnings about constructions that some users consider questionable, and that are easy to avoid (or modify to prevent the warning), even in conjunction with macros.
```

alexhultman:
@ksergey You are missing my point entirely here. The point I'm trying to make is, we all have different **opinions**.

**I'm** perfectly fine with some harmless warnings, **you** are not fine with unused variable warnings but fine with missed loop-optimization warnings. **Others** might require loop optimizing warnings to be fixed, what do I know. We all have different opinions and arguing about these things is as productive as arguing whether to use 2 or 4 spaces as indentation or a tab. It all comes down to preferences and we all have different ones. This PR is probably the most unnecessary one in a long time because it doesn't solve any kind of problem in any way.

ksergey:
@alexhultman
If warning is harmless why not mask it? 
Is it not annoying to search new warnings in compiler output after code change? Or you silently ignore all warnings?

alexhultman:
I run g++ *.cpp > /dev/null and commit whatever change I made, yes

alexhultman:
https://github.com/uNetworking/uWebSockets/blob/master/tests/uWebSockets.pro#L34

-Wno-unused-parameter is your friend

ksergey:
`-Wno-unused-parameter`  
It's not harmless warning

https://stackoverflow.com/questions/308277/what-are-the-consequences-of-ignoring-warning-unused-parameter

Looks like you never written production quality code. It's bad practice to ignore warnings.
If warning false-positive it should be masked in code.

Looks like this dispute is pointless. Let's stop it.

alexhultman:
@ksergey You see that little button in the top right? Called "Fork", you can click it and be the production ready hero the world deserves. In the mean time I will focus on making people consume less hardware resources.

alexhultman:
I found you a costume: https://www.simplyfancydress.co.uk/adult-captain-y-fronts-super-hero-costume~63428/

Btw, this shop is not production ready because it makes use of µWS so I wouldn't trust it.

PR #741, opened 2017-11-16
added explanation why it doesnt work on OS X

bydga:
Since your issues are fuckin closed, this is the only way to contact you. The websockets doesn't work on High Sierra, however it connects and works on older Sierra. Maybe instead of bitching around - http://prntscr.com/hb7kvi <-- I had to search for this like an idiot in the google cache, you could ask for more details (when the ppl are unable to provide you the OS/compiler versions etc themselves).
I dont expect you to fix it - just provide this info to the people so they wont have to search for it like I did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 <div align="center"><img src="misc/images/logo.png"/></div>
 
+# Module not working on new versions of OS X.
+Since your issues are fuckin closed, this is the only way to contact you. The websockets doesn't work on High Sierra, however it connects and works on older Sierra. Maybe instead of bitching around - http://prntscr.com/hb7kvi <-- I had to search for this like an idiot in the google cache, you could ask for more details (when the ppl are unable to provide you the OS/compiler versions etc themselves).
+I dont expect you to fix it - just provide this info to the people so they wont have to search for it like I did.
+
+
+
 `µWS` is one of the most lightweight, efficient & scalable WebSocket & HTTP server implementations available. It features an easy-to-use, fully async object-oriented interface and scales to millions of connections using only a fraction of memory compared to the competition. While performance and scalability are two of our top priorities, we consider security, stability and standards compliance paramount. License is zlib/libpng (very permissive & suits commercial applications).
 
 


ramonberrutti:
I hope you get banned. Github allow you to fork the proyect and start to fix. The new programmers want that all the code be copy-paste. I never see the word: "thanks" to a developer in github

alexhultman:
Where is the explanation? I'm heavily doubting a system ipdate would break existing programs. 99% chance you messed up the build like most others.

jmealo:
@alexhultman: an OSX upgrade typically breaks things. I've had homebrew cause a delay of 8-12 hours while the upgrade process failed repeatedly to move files. It has gotten better but often corrective steps need to be taken on upgraded machines versus clean installs. This is common/well-known and one of the reasons I stay a release behind for OSX. It'd be helpful if people could collaboratively troubleshoot these issues in an issue tracker as OSX environmental issues are hard to pin down and are not for the faint of heart. This can mostly be avoided by never linking against system libraries that change between releases and using a package manager. In the past, this was not enough as the update process borked homebrew.

... Or you could just claim superiority and belittle members of your community... as per usual.

jmealo:
FWIW, @bydga cursing out @alexhultman on his own repo only justifies his poor behavior, you're both acting like children, but this is his house. The sarcasm in a PR is uncalled for. You could have spent this time troubleshooting the issue or writing a serious note linking to an external forum to discuss the issues that you're having.

alexhultman:
I'm guessing you all are Node.js people, right? The code does not rot into oblivion just because of an update. This code is written against interfaces from the 80s, 90s and 00s. These things are supported forever. You are obviously building it wrong. You mention Homebrew breaking and how users link to wrong libraries - again, go fix your own problems because these are build problems. I will take the time to extra check when I have time for it, but so far EVERY build error report has been due to invompetence.

alexhultman:
@bydga I just upgraded my installation to High Sierra and ran µWS. I ran both an old executable, and then recompiled it from source on High Sierra. Both executables worked and I got my messages echoed.

As much as I like to offend people, and this would be the perfect opportunity, I just realized that I have absolutely no time for you and that you mean nothing to me. So I will instead save my focus for something productive, and I'll just wish you good luck in life and a nice week :heart:

PR #742, opened 2017-11-19
Strenghten TLS cipher selection on secure connection

dvergeylen:
Context creation (Networking.cpp) already restricted connections to be TLSv1, TLSv1.1 or TLSv1.2 only. However, not specifying a DH params file forbids openssl to choose among DH ciphers.

As stated in `man SSL_CTX_set_cipher_list`:
>[...] On the server side, additional restrictions apply. All ciphers have additional requirements. ADH ciphers don't need a certificate, but DH-parameters must have been set. All other ciphers need a corresponding certificate and key.

This PR allows a user to specify a DH params file which will allow to restrict ciphers to "strong ones only". If no DH params file is provided, cipher list stays as is (TLSv\* only).

I spotted that Node.cpp also creates SSL contexts but I would prefer to have your thoughts on this before going further.


alexhultman:
Thanks for the PR. OpenSSL is a huge library of a lot of things and I think this PR is too specific to be useful by the masses. If you want specific features you can just subclass the SSLContext and make your changes to the native OpenSSL context. It would be a slippery slope to go down to try and wrap everything of OpenSSL instead of just exposing OpenSSL itself. The same applied to another PR where someone needed some other specific feature I cannot remember -> better to go native OpenSSL in those cases.

dvergeylen:
Thank you for your reply.

>OpenSSL is a huge library of a lot of things

Hell Yeah! 😅

By `SSLContext` you mean `uS::TLS::Context` class?

PR #743, opened 2017-11-22
verifyClient

cuteyax:
Hi there,
Is there any way to pass data from verifyClient to connection event?


alexhultman:
What do you want to merge?

PR #744, opened 2017-11-27
Added static library build

izaid:
This modifies the Makefile, in a very minimal way, to also build libuWS as a static library. I have tested this on Linux and OS X, it works.


lkramer:
I like this, but you should make the static library is also copied when calling make install.

alexhultman:
Yeah you could compile once, link twice (shared object / archive). Probably good to also delete generated object files after and also install libuWS.so and libuWS.a like mentioned

izaid:
Okay, cool, I'll update the PR when I get a chance.

PR #745, opened 2017-12-03
Instructions for installing libuv on Debian|Ubuntu

alexslipknot:


alexhultman:
Why is there no zlib or ssl packages? And why libuv1 when libuv1-dev has this as dependency. So many questions

alexslipknot:
@alexhultman indeed. Please excuse me. I just want to inform you that `libuv` does not have candidate to install. So on Debian/Ubuntu you have to install `libuv1`

alexhultman:
Shouldn't it be -dev packages?

alexslipknot:
To be sure I did `make` on clean fresh installed Ubuntu. So I found that `zlib` (`zlib1g-dev`) library is already installed. So I removed it. Also I found that on fresh Ubuntu `libssl-dev` is absent . So I've added it. Thanks for the tip.

alexhultman:
I'm not feeling this PR, thanks

PR #746, opened 2017-12-09
Fix test crashes in Linux and able to skip certain tests using arguments

chkuang-g:
Hi Alex,

Thank you for creating this library. I'd like to utilize this test script in our build system.  However, I encountered some issues.

1. testClosing() always causes Segmentation Fault in Linux because the captured reference is destroyed.
2. Certain tests are not allowed in our build system (ex. connecting to external server in testConnection) and I'd like to be able to skip some tests through arguments.

This pull request is basically to address these two issues.  The default behavior remains the same.

Please review and let me know if you have any questions.

Thank you,
Shawn

### DIFF
--- a/tests/main.cpp
+++ b/tests/main.cpp
@@ -322,12 +322,17 @@ void testClosing() {
     uWS::Hub h;
     const char *closeMessage = "Closing you down!";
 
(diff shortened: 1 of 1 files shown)


chkuang-g:
Hi Alex,
I have merge all your latest changes.  Could you take a look and see if this is useful?

Thank you!

PR #747, opened 2017-12-11
Postinstall cleanup

wision:
Hello.. thanks for the module. I wasn't able to find package.json file in this repo, hence the empty commit, but have you considered using something like:
`find -name "uws_*.node" -a -not -name "*$(echo "console.log(process.platform + '_' + process.versions.modules)" | node)*" -delete`
as postinstall script, so the uws module does not keep binaries from other OS and nodejs versions modules after the installation? We use uws in well over hundred of applications deployed in thousands of instances and making this module smaller in size, would help us keep our application images lighter. Thanks for your reply.


ramonberrutti:
Hi, first don't use the pull requests for Support, Second for binding library go to: https://github.com/uNetworking/bindings this is only the main library.

wision:
Sorry, but I didn't know any other way of publicly contacting the maintainer of uws nodejs module, other than a pull request here. Anyway, thanks for the link.. I didn't know about it, since npmjs.org redirects here as a source of uws module. https://www.npmjs.com/package/uws

alexhultman:
It is a good idea but are you sure this is cross platform?

wision:
I will test it and send pull request to bindings repository.

PR #749, opened 2017-12-18
Use '> /dev/null 2>&1' redirection

otamachan:
Hi,

For Debian/Ubuntu, ``dash`` is used as ``/bin/sh`` (https://wiki.ubuntu.com/DashAsBinSh) and `dash` does not support ``&>`` style redirection(https://wiki.ubuntu.com/DashAsBinSh#A.26.3E). I tried to change the ``SHELL`` environmental variable to use ``/bin/bash``, but ``popen`` does not seem to use that variable.
This PR changes the redirection style.

I tested by the following Dockerfile

```
FROM ubuntu:16.04
#FROM debian:jessie
RUN apt-get update && \
    apt-get install -y build-essential libssl-dev netcat-openbsd
ADD . /uWebSockets
RUN cd uWebSockets && make && make install && g++ -std=c++11 -O3 tests/main.cpp -Isrc -o testsBin -lpthread -L. -luWS -lssl -lcrypto -lz
ENV SHELL /bin/bash
CMD cd uWebSockets && ./testsBin
```

```bash
$ docker build -t test .
$ docker run --rm --net=host test
```

Unfortunately ``debian:jessie`` failed with segmentation fault. It might be related to #746 .

Thanks.

### DIFF
--- a/tests/main.cpp
+++ b/tests/main.cpp
@@ -828,44 +828,44 @@ void testHTTP() {
         FILE *nc;
 
         // invalid data
-        nc = popen("nc localhost 3000 &> /dev/null", "w");
+        nc = popen("nc localhost 3000 > /dev/null 2>&1", "w");
         fputs("invalid http", nc);
         pclose(nc);
 
-        nc = popen("nc localhost 3000 &> /dev/null", "w");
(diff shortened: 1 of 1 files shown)


otamachan:
Sorry, duplicated #704.

PR #752, opened 2018-01-09
Fixing implicit conversion warnings.

jacobsologub:


src/WebSocketProtocol.h
```diff
@@ -148,7 +148,7 @@ class WIN32_EXPORT WebSocketProtocol {
         } else {
             wState->state.spillLength = 0;
             wState->state.wantsHead = false;
-            wState->remainingBytes = payLength - length + MESSAGE_HEADER;
+            wState->remainingBytes = (unsigned int) payLength - length + MESSAGE_HEADER;
```
alexhultman: This causes a bug

alexhultman: You need parenthesis here otherwise you mess up the logic

src/Networking.h
```diff
@@ -216,7 +216,7 @@ struct WIN32_EXPORT NodeData {
     static void asyncCallback(Async *async);
 
     static int getMemoryBlockIndex(size_t length) {
-        return (length >> 4) + bool(length & 15);
+        return (int) (length >> 4) + bool(length & 15);
```
alexhultman: Same here, you change the logic.

jacobsologub:
Fixed

PR #753, opened 2018-01-10
Small fix for Epoll version of Async

chkuang-g:
In Epoll version of Async::close(), delete itself as Async* instead of Poll*


src/Epoll.h
```diff
@@ -243,7 +243,7 @@ struct Async : Poll {
         Poll::stop(loop);
         ::close(state.fd);
         Poll::close(loop, [](Poll *p) {
-            delete p;
+            delete (Async*) p;
```
alexhultman: Star to the right

PR #754, opened 2018-01-10
Fix using as dynamic library.

mohe2015:


ramonberrutti:
why add ssl, when you can add in your main program :)

mohe2015:
@ramonberrutti Yeah, I realized that my ssl installation was buggy. Therefore it didn't work to add it to the main program (using build system).

alexhultman:
I think this is a good idea. I didn't realize this was possible but of course it is.

```
[alexhultman@localhost uWebSockets]$ ldd libuWS.so 
	linux-vdso.so.1 (0x00007ffee9d87000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007feb63a9a000)
	libm.so.6 => /lib64/libm.so.6 (0x00007feb63745000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007feb6352e000)
	libc.so.6 => /lib64/libc.so.6 (0x00007feb63149000)
	/lib64/ld-linux-x86-64.so.2 (0x00007feb6404c000)
```

vs.

```
[alexhultman@localhost uWebSockets]$ ldd libuWS.so 
	linux-vdso.so.1 (0x00007ffeb9514000)
	libssl.so.1.1 => /lib64/libssl.so.1.1 (0x00007f8b9f903000)
	libcrypto.so.1.1 => /lib64/libcrypto.so.1.1 (0x00007f8b9f47b000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f8b9f0f5000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f8b9eda0000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f8b9eb89000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f8b9e7a4000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f8b9fd9b000)
	libz.so.1 => /lib64/libz.so.1 (0x00007f8b9e58d000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f8b9e389000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f8b9e16a000)
```

Then you just link to libuws.so and be done (maybe time to lowercase it from libuWS.so as well?)

alexhultman:
Ah, I already tried this long before but it doesn't seem to work. Some headers include unique references to inflateEnd so you still need to manually link with -lz. If headers only held declarations it would work I guess but now they hold implementation as well (templated lib).

I get this:

```
[alexhultman@localhost examples]$ g++ echo.cpp -luWS
/usr/bin/ld: /tmp/ccTlkIrg.o: undefined reference to symbol 'inflateEnd'
//usr/lib64/libz.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

Which is better than before but still not perfect. I need to go full `g++ echo.cpp -luWS -lz` for it to compile. I also suspect you still need to manually link with ssl if you would use any of the SSL features of the lib, as they also live in headers.

mohe2015:
Yes, that was the reason I closed it again. It didn't really work like I wanted.

PR #758, opened 2018-01-28
Make members accessible in derived class

otamachan:
Hi,

I want to use uWebSockets as a websocket client from behind a proxy.
To implement this, I overrided [``connect``](https://github.com/otamachan/uWS_proxy/blob/2454ff8cc30bbfe77dab90dc580f36a269524944/src/ProxyHub.cpp#L102) method of ``Hub`` class and made a [``ProxyHttpSocket``](https://github.com/otamachan/uWS_proxy/blob/8f9e1a5e15737839b7b86eeaa391b2aa9d8f784e/src/ProxyHTTPSocket.h) class derived from ``HttpSocket`` class to communicate with a proxy server to establish a connection. Unfortunately  some of the members are not accessible in a derived class. This PR makes them accessible.

A full sample is [here](https://github.com/otamachan/uWS_proxy).

If you know a better way to use uWebSockets behind a proxy, please let me know.

Thanks in advance.

### DIFF
--- a/src/Node.h
+++ b/src/Node.h
@@ -13,7 +13,7 @@ enum ListenOptions : int {
 };
 
 class WIN32_EXPORT Node {
-private:
+protected:
     template <void C(Socket *p, bool error)>
     static void connect_cb(Poll *p, int status, int events) {
         C((Socket *) p, status < 0);
@@ -71,7 +71,6 @@ private:
         } while ((clientFd = netContext->acceptSocket(serverFd)) != INVALID_SOCKET);
     }
 
-protected:
     Loop *loop;
     NodeData *nodeData;
     std::recursive_mutex asyncMutex;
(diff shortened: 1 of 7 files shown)


alexhultman:
I don't get what you are doing, checked the sources and they look very "below the surface" but anyways, protected is probably better than private

alexhultman:
This only changes HTTPSocket, Hub and Group. Either everything is private or everything is protected

otamachan:
Thanks for your quick response. I will change everything is protected.

otamachan:
Done.

I kept these private. 
https://github.com/uNetworking/uWebSockets/blob/d990bef9f2dc483d235cb4bccc357b4240ec2c51/src/Extensions.cpp#L14
https://github.com/uNetworking/uWebSockets/blob/d990bef9f2dc483d235cb4bccc357b4240ec2c51/src/Libuv.h#L88

src/Node.h
```diff
@@ -13,7 +13,7 @@ enum ListenOptions : int {
 };
 
 class WIN32_EXPORT Node {
-private:
+ protected:
```
alexhultman: space

otamachan: Fixed

PR #760, opened 2018-01-29
redesign Makefile

rindeal:


alexhultman:
If you can make it smaller that would be great. Currently this change makes it larger which is not prioritized

PR #762, opened 2018-02-01
Support external event loops via non-blocking poll() alt. to run()

jamesbeilby:
For use in applications that already have a primary event loop, so they cannot call blocking run(). They may instead call poll() within this loop to delegate to uWebSocket's secondary event loop (whether Asio, libuv or epoll). It is assumed that this will be called very frequently, e.g. "busy spin".

### DIFF
--- a/src/Asio.h
+++ b/src/Asio.h
@@ -22,6 +22,10 @@ struct Loop : boost::asio::io_service {
     void run() {
         boost::asio::io_service::run();
     }
+
+    void poll() {
+    	boost::asio::io_service::poll();
+    }
 };
 
 struct Timer {
--- a/src/Asio.h
+++ b/src/Asio.h
@@ -24,7 +24,7 @@ struct Loop : boost::asio::io_service {
     }
 
     void poll() {
-    	boost::asio::io_service::poll();
+        boost::asio::io_service::poll();
     }
 };
 
--- a/src/Epoll.cpp
+++ b/src/Epoll.cpp
@@ -58,6 +58,8 @@ void Loop::doEpoll(int epollTimeout) {
 }
 
 void Loop::run() {
+    // updated for consistency with libuv impl. behaviour
+    timepoint = std::chrono::system_clock::now();
     while (numPolls) {
         doEpoll(delay);
     }
@@ -66,6 +68,9 @@ void Loop::run() {
 void Loop::poll() {
     if (numPolls) {
         doEpoll(0);
+    } else {
+        // updated for consistency with libuv impl. behaviour
+        timepoint = std::chrono::system_clock::now();
     }
 }
 
--- a/src/Epoll.cpp
+++ b/src/Epoll.cpp
@@ -9,55 +9,64 @@ std::recursive_mutex cbMutex;
 void (*callbacks[16])(Poll *, int, int);
 int cbHead = 0;
 
-void Loop::run() {
-    timepoint = std::chrono::system_clock::now();
-    while (numPolls) {
-        for (std::pair<Poll *, void (*)(Poll *)> c : closing) {
-            numPolls--;
+void Loop::doEpoll(int epollTimeout) {
+    for (std::pair<Poll *, void (*)(Poll *)> c : closing) {
+        numPolls--;
 
-            c.second(c.first);
+        c.second(c.first);
 
-            if (!numPolls) {
-                closing.clear();
-                return;
-            }
-        }
-        closing.clear();
-
-        int numFdReady = epoll_wait(epfd, readyEvents, 1024, delay);
-        timepoint = std::chrono::system_clock::now();
-
-        if (preCb) {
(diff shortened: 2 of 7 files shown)


alexhultman:
What kind of loop do you have in mind?

src/Node.cpp
```diff
@@ -64,6 +64,10 @@ void Node::run() {
     loop->run();
 }
 
+void Node::poll() {
```
alexhultman: makes sense

src/Libuv.h
```diff
@@ -24,6 +24,10 @@ struct Loop : uv_loop_t {
     void run() {
         uv_run(this, UV_RUN_DEFAULT);
     }
+
+    void poll() {
```
alexhultman: makes sense

src/Asio.h
```diff
@@ -22,6 +22,10 @@ struct Loop : boost::asio::io_service {
     void run() {
         boost::asio::io_service::run();
     }
+
+    void poll() {
```
alexhultman: makes sense

src/Epoll.cpp
```diff
@@ -9,55 +9,64 @@ std::recursive_mutex cbMutex;
...
+
+void Loop::run() {
+    while (numPolls) {
+    	doEpoll(delay);
+    }
+}
+
+void Loop::poll() {
+	if (numPolls) {
+		doEpoll(0);
```
alexhultman: tabs here

src/Epoll.cpp
```diff
@@ -9,55 +9,64 @@ std::recursive_mutex cbMutex;
...
     }
+
+    if (postCb) {
+        postCb(postCbData);
+    }
+}
+
+void Loop::run() {
+    while (numPolls) {
+    	doEpoll(delay);
```
alexhultman: spaces and tabs here

src/Epoll.cpp
```diff
@@ -9,55 +9,64 @@ std::recursive_mutex cbMutex;
 void (*callbacks[16])(Poll *, int, int);
 int cbHead = 0;
 
-void Loop::run() {
-    timepoint = std::chrono::system_clock::now();
```
alexhultman: Why did you remove this? It breaks timers in some cases

jamesbeilby:
Thanks Alex.

For context, I have a low latency application following Actor pattern (with message passing between threads), and this Actor framework uses a custom dispatch/event loop as it needs to support non-standard I/O (like user-space network sockets) - so ruling out epoll for example. It is useful to be able to handle HTTP/WS requests on one of these Actor threads, as the same thread can then send/receive messages to the rest of the application, without any synchronization.

I have fixed the tabs.

I was going to ask about timepoint -- I thought it possible that it was set conservatively, noting that it is updated prior to invoking callbacks (line 26). All the timers I could find are created either from within these callbacks or possibly before run() is called. Perhaps this is to handle case where numPolls == 0 when run is called, or I am just missing the point?

Cheers

alexhultman:
> Perhaps this is to handle case where numPolls == 0 when run is called, or I am just missing the point?

Yep. It comes from strange libuv behavior:

http://docs.libuv.org/en/v1.x/loop.html#c.uv_update_time

Calling Hub::run() with no polls should update the loop's timepoint like is done in libuv to keep same behavior.

jamesbeilby:
Makes sense thanks. Have reverted this.

chronoxor:
Is it possible provide Hub/Node initialization with external Loop? For example I have my own asio::io_service and I want to use it with uWebSockets Hub.

jamesbeilby:
chrono, it would be possible with this change to have your own io_service and uWebSockets' io_service running on the same thread, but this would be inefficient. I would suggest maybe a separate change that handles "defaultLoop = true" in Loop::Loop in Asio.h, or that extends defaultLoop to allow a custom loop to be constructed and passed.

alexhultman:
Yes the idea is to support seamless integration with libuv & asio projects. You could possibly get the loop from µWS using Hub::getLoop() but otherwise you need to PR that change so that you can pass in any io_service (probably just make that a void ptr or something).

chronoxor:
I'll try to create such functionality with external loop initialization after I finish with Asio Standalone loop, which works without boost dependency.

Injecting into the static createLoop() is a bad idia, because it is difficult to control workflow especially in multithread env.

chronoxor:
Suddenly I found that uWebSockets lib does not support ASIO on Windows platform because the lack of boost::asio::posix::stream_descriptor.

Then I tried to replace it with boost::asio::ip::tcp::socket. Some modifications were made in Asio.h (see attachment). And I managed to compile and run echo.cpp sample. But when I connect to it with a client WebSocket from my Chrome I get Access Violation error in debugger.

[Asio.h.zip](https://github.com/uNetworking/uWebSockets/files/1700345/Asio.h.zip)
uwebsockets

PR #763, opened 2018-02-09
Templatize uS::Timer::start to accept lambdas (Linux, gcc 5.4)

manutalcual:
At least with Linux, GCC 5.4, the Timer::start function doesn't accept a lamda parameter. This change fixes that.

### DIFF
--- a/src/Epoll.h
+++ b/src/Epoll.h
@@ -77,7 +77,8 @@ struct Timer {
         this->loop = loop;
     }
 
-    void start(void (*cb)(Timer *), int timeout, int repeat) {
+	template<typename Func>
+    void start(Func cb, int timeout, int repeat) {
         loop->timepoint = std::chrono::system_clock::now();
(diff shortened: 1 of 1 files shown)


alexhultman:
This change is incomplete, badly indented and does not fix anything. You just postpone the conversion error from Timer::start to Timepoint construction - you are still dealing with function pointers and not lambdas. A lambda is convertible to a function pointer if it captures no data, but not otherwise.

I get where you want to move things and it makes sense from a C++ standpoint but this lib has legacy design from libuv, which is a C library where lambdas doesn't exist. Use the userData and the function pointer to achieve the same thing. It does make a good change if the lib would take a more C++-ish design later on but like mentioned this change makes no sense and is incomplete.

manutalcual:
Alex,

Yes. Sorry. I've seen tests/main.cpp receiving a lambda and didn't realize
it isn't capturing anything.
Thanks for your answer.

Kind regards,
Manu

2018-02-09 10:54 GMT+01:00 Alex Hultman <notifications@github.com>:

PR #765, opened 2018-02-18
redesign Makefile

rindeal:
This is a followup of #760. I cut it down by 16 SLOC and it is now at reasonable size, reasonable readability and reasonable functionality. There are just 5 blank lines which could be nuked eventually at the cost of decreased readability. There are also no targets specialized for some platforms anymore, everything is unified with minimal branching for Darwin.

Abilities gained:

- override default CXXFLAGS, useful for distro packagers
- specify own CXXLD, {CXX,LD}FLAGS, LDLIBS, RM
- specify own library names, useful for someone who wants to specify SONAME, like Debian packagers
- build static library
- build object files separately from linking/archiving
- ability to use jobserver
- faster build time thanks to parallel compilation
- test binary gets cleaned

Bugs fixed:

- CFLAGS is no longer used for CXX

Other PRs closed:

- https://github.com/uNetworking/uWebSockets/pull/744

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,33 @@
-osx_image: xcode7.3
-os:
-  - osx
-script: brew install --force openssl libuv && make && make install && make tests && time ./testsBin
+language: cpp
+
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+      sudo: false
+      addons:
+        apt:
+          packages:
+            - tree
+            - libuv-dev
+    - os: osx
+      osx_image: xcode7.3
+
+install:
+  - 'if [[ "${TRAVIS_OS_NAME}" == "osx" ]] ; then brew update && brew install --force openssl libuv tree ; fi'
+
+script:
+  - 'set -e'
+  - 'export PREFIX="/tmp/install"'
+  - 'export MAKEFLAGS="-j8"'
+
+  - 'make'
+
+  - 'make install'
+  - 'tree "${PREFIX}"'
+
+  - 'make tests'
+  - 'time $([[ "${TRAVIS_OS_NAME}" == "linux" ]] && echo "catchsegv") ./testsBin'
+
+notifications:
+    email: false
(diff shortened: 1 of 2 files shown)


alexhultman:
Looks nice indeed. Will check it out in the morning

alexhultman:
It is not building in parallel and it is not working on OS X. If you want things merged you need to provide working changes.

rindeal:
It's building in parallel if you tell `make` to do so, like by specifying the `--jobs` option.

rindeal:
And what exactly is not working for you on OSX. The only difference between then and now was the addition of `-s` linker option for Darwin, which I fixed by now. Otherwise the commands are exactly the same. If you have still some errors now, then you will have to post some debugging information like the error messages it spits out etc.

alexhultman:
I don't have to do anything. I'm not going to merge a PR that breaks one platform because of your lacking testing. I'm not your baby sitter I don't wipe peoples assholes for free. Fix, test and update your changes or they won't be merged.

trivigy:
@alexhultman when are you planning on merging and releasing this?

rindeal:
@alexhultman By doing an extensive testing I've found some issues with the handling of the Darwin platform in my Makefile and fixed them. It now compiles libraries and tests on both Linux and Darwin platforms, but tests do not pass on Linux, which I deemed to not be a problem of my Makefile. I have also attached the Travis-CI config file I used to test my Makefile with.

alexhultman:
I'm still trying to consume this Makefile piece by piece. Why did you change it to a debug build? It's not intended to be a debug build. And what is CPPFLAGS vs. CXXFLAGS? I don't even get the half of what this is supposed to be, and why?

alexhultman:
Seriously. The old Makefile builds a 183 kb large shared object. This "remake" builds a 2.1 MEGABYTE of debug build what do you expect me to do with this? Why are you mixing "remake" with actual changes to the build?

alexhultman:
These flags are not drunkenly put there by mistake. I was not consuming large amounts of Absinthe when doing this. What is your point in this exactly?

rindeal:
`CPPFLAGS` are flags for the C preprocessor, `CXXFLAGS` are flags for the C++ compiler, although they're the same program and build step in this case, they're exported separately during package builds in my Linux distro.

The debugging symbols are there by default so that you get a nice backtrace when your tests segfault. If you don't want them to be present in your setup run make like this: `CXXFLAGS="-O3" LDFLAGS="" make ...`.

Another important change is the switch from weak linking to a hard one since weak linking is not a wise thing to do for a shared library without pkg-config files specifying its deps.

The whole point of this was to become able to build this software with the package manager of my choice primarily by making the Makefile standards-compliant.

PR #766, opened 2018-02-20
Added CMakeLists.txt to build uWS static library

bsergeev:
CMake facilitates the use of your awesome _uWebSokets_ library in other CMake driven projects by simply including its location:
```
add_subdirectory(${UWS_LOCATION})
```
A complete example of such CMake file can be found here:
https://github.com/bsergeev/StrawPoll/blob/master/CMakeLists.txt

The use of CMake makes bundling Visual Studio projects obsolete.


alexhultman:
This project prefers Unix systems and its standard make utility.

bsergeev:
My only point is that multi-platform nature of your project would benefit from multi-platform build tool like CMake... This makes it easier for other people to use your library without forking it.

I build my projects in Windows 10 and in Ubuntu running in WSL (Windows Subsystem for Linux) from he same source files using the same CMake files and build script, e.g.:
```
if [ "$OS_NAME" = "Linux" ]; then
  cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=$BUILD_CONFIGURATION ..
  make -j4
elif [ "$OS_NAME" = "Windows" ]; then
  cmake -G "Visual Studio 15 2017 Win64" ..
  cmake --build . --target ALL_BUILD --config $BUILD_CONFIGURATION
fi
```

Any way, do as you wish with my PR.
And thank you for the great library! 👍

JulZimmermann:
I'd love CMake support!

As bsergeev wrote, CMake support would make the VC++ projekt file obsolete, as Visual Studio can handle CMake projects in its newest versions and .sln files can be generated for older versions.

I (and many other people) use CMake for modern, platform independent C++ development

alexhultman:
CMake is not used here

loint:
Makefile is more popular, normally developer no need to install CMake if they do not use CMake from the beginning. Thank you for your strictly to make things stable @alexhultman .

PR #767, opened 2018-02-24
Add support for CMake

selavy:
If you would like to use it, here is a cmake build implementation.  It is annoying to add another dependency, but cmake is sort-of the leader in the C++ build space so people should have it.

I am getting a slightly large .so.  Based on output from nm, I believe both libs have the same symbols.   So my guess is that because cmake is doing an incremental build, whereas the Makefile passes all .cpp in together, it is getting a little better optimization.  I threw in LTO support, but that still doesn't get the size below what the Makefile generates.  If you want to check the flags being used for the build, you can do "make VERBOSE=1".

-----------------------------------------------------------------------------------------------------

Uses c++ to build libuWS library as well as the tests.

Basic Usage (Linux Makefiles):
    $ mkdir build  # out-of-source builds are encouraged
    $ cd build/
    $ cmake ..
    $ make -j
    $ make install

Windows Project Generation:
    $ mkdir build
    $ cd build/
    $ cmake -G "Visual Studio Visual Studio 15 2017 Win64" ..

XCode Project Generation:
    $ mkdir build
    $ cd build/
    $ cmake -G "Xcode" ..

Other generators are listed here: https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html.

REVISIT: On Linux, switch to Ninja as default?

Other options:
    +CMAKE_BUILD_TYPE={Debug, Release, MinSizeRel}
    +BUILD_SHARED_LIBS={ON,OFF}
    +USE_LTO={ON,OFF}

    e.g. $ cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=OFF -DUSE_LTO=ON ..

clang 3.8 is complaining about the "-s" strip flag, so it only gets
added if the compiler is gcc.

TODO: check on strip support on Windows

TODO: setup cpack and ctest

Note: I do not have access to OS X to test, but cmake *should* just
work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+build/
+*.so
+testsBin/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,75 @@
+cmake_minimum_required(VERSION 3.0)
+project(uWS)
+
+SET(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+
+# Default to shared libs if not set
+if (NOT DEFINED BUILD_SHARED_LIBS)
+    set(BUILD_SHARED_LIBS ON)
+endif()
+if (BUILD_SHARED_LIBS)
+    message("-- building shared libs")
+else()
+    message("-- building static libs")
+endif()
+
+# REVISIT: maybe should default to Debug, and give installation
+#          notes for building release?
+# Default to release build if not set
+if ("${CMAKE_BUILD_TYPE}" STREQUAL "")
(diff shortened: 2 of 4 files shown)


selavy:
Never mind, I missed that you already said you don't want cmake.

alexhultman:
> but cmake is sort-of the leader in the C++ build space so people should have it.

I know people who will have a fist fight with you over what build system is best. That's why I refuse to have any other than the most basic one that has been there for 30 years now and still does the job. Even when this project had CMake (oh yes it did have) people HATED IT and IT was the most PR:ed piece of the entire project. You see my view now?

PR #770, opened 2018-03-24
Fix SSL library init order

Matheus28:
It was possible for it to not have been initialized yet if we're calling createContext in the initialization of a global variable in another file, since there's nothing that guarantees the initialization order between global variables in different files.

There's also another init further down in this file related to Windows, but I haven't been using uWebSockets on Windows that much to hit an issue with that yet.

### DIFF
--- a/src/Networking.cpp
+++ b/src/Networking.cpp
@@ -27,13 +27,13 @@ Context::~Context()
     }
 }
 
-struct Init {
-    Init() {SSL_library_init();}
-    ~Init() {/*EVP_cleanup();*/}
-} init;
-
 Context createContext(std::string certChainFileName, std::string keyFileName, std::string keyFilePassword)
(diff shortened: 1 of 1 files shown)


alexhultman:
What version of openssl? What platform? Openssl 1.1 deprecated this function actually, it does init by itself now

Matheus28:
This is on macOS:

`$ openssl version
LibreSSL 2.2.7`

I'm pretty sure that's also the one that's getting linked, but I can double check if you'd like.

I noticed this problem and debugged it down to that `SSL_CTX_new(SSLv23_server_method())` call failing. Moving the init there makes sure it's initialized before the function continues, which turned out to be the issue.

alexhultman:
That's definitely prior to openssl 1.1 ;p

PR #771, opened 2018-03-27
loop through the list returned by getaddrinfo

witaly-iwanow:
getaddrinfo returns a linked list, you should loop through it and try entries until you succeed:
http://beej.us/guide/bgnet/html/multi/syscalls.html#getaddrinfo

Here's an example of doing it properly. Curl tries to connect to localhost, getaddrinfo returns the list with ::1 as the first entry, connection to it fails and curl proceeds to the next entry which is good old IPv4 127.0.0.1 and it succeeds:
```
curl -v localhost:8000
* Rebuilt URL to: localhost:8000/
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 8000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8000 (#0)
...
```

### DIFF
--- a/src/Networking.cpp
+++ b/src/Networking.cpp
@@ -27,13 +27,13 @@ Context::~Context()
     }
 }
 
-struct Init {
-    Init() {SSL_library_init();}
-    ~Init() {/*EVP_cleanup();*/}
-} init;
-
 Context createContext(std::string certChainFileName, std::string keyFileName, std::string keyFilePassword)
 {
+    static struct Init {
+        Init() {SSL_library_init();}
+        ~Init() {/*EVP_cleanup();*/}
+    } init;
+    
     Context context(SSL_CTX_new(SSLv23_server_method()));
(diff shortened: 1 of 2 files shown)


alexhultman:
Client connection is lacking currently, you also need to do those calls in a separate thread to keep it "async" as it currently can block if DNS resolution takes time.

witaly-iwanow:
Not sure it's the right place to discuss that SSL init change, but anyway, I experienced the same behaviour - it did cause a crash on macOS and Windows in my set up (my own code + uWS source + uv source + static ssl and crypto, all linked together into an executable without any external dependencies)

alexhultman:
It is reset now

alexhultman:
All per-platform code will be moved to uSockets repo, you can tinker with that repo if you want

PR #773, opened 2018-05-05
Shutdown HttpSocket immediately on end()

not-implemented:
When answering a HttpRequest with `res->end()`, the connection is currently not closed immediately but 1-2 secs later by the timeout in `Group::addHttpSocket()`. Normally this works fine, but when sending custom headers (with `res->hasHead = true;`) without a Content-Length-Header, the HTTP client has to wait until the server closes the connection.

For some simple (error-)responses it's unhandy to calculate the Content-Length. Wouldn't it be better, that `end()` always closes the connection immediately after the data has been sent by default?

I currently don't have an idea where to implement this. And I currently don't understand what's the difference between `HttpResponse::write()` and `HttpResponse::end()` as it is implemented completely different. Intuitively I would think, `write()` does the same as `end()` but `end()` closes the connection afterwards, or am I completely wrong here?

```
    h.onHttpRequest([&](HttpResponse *res, HttpRequest req, char *data, size_t length,
                        size_t remainingBytes) {
        res->hasHead = true;
        res->end(response.data(), response.length());
    });
```

BTW: I loved to read "if v0.15 ever happens HTTP will be a major focus." ;-)

### DIFF
--- a/src/HTTPSocket.h
+++ b/src/HTTPSocket.h
@@ -264,6 +264,8 @@ struct HttpResponse {
             }
 
             freeResponse(httpSocket);
+
+            // TODO: httpSocket->shutdown() when finished sending?
         }
     }
 


alexhultman:
This is the behavior of http 1.1 - closing the connection is more of a http 1.0 kind of thing. One thing is certain: http support is very limited and very awkward in 0.14

not-implemented:
No problem ;-) ... I use it just for some error responses and as status-interface for the WebSocket server.

Would it be safe, to call `res->getHttpSocket()->shutdown();` directly after `res->end();` as a workaround? Or is the only safe way currently to generate a Content-Length header in the response?

alexhultman:
I don't remember, I think it should be possible to shutdown yourself

not-implemented:
Ok, let's see ... thanks.

PR #774, opened 2018-05-06
Support OpenSSL 1.1.0 and TLS for networking

FiniteReality:
~~This effectively drops support for OpenSSL 1.0.0 since it uses stuff only added in OpenSSL 1.1.0, and should keep behaviour almost the same as before.~~ EDIT: Uses a compile-time macro to decide whether to use OpenSSL 1.0.0-compatible APIs or OpenSSL 1.1.0-compatible APIs.

With this change, SSLv2 support is lost when compiling against OpenSSL 1.1.0 as it was removed, but I don't think that's a large enough issue to worry about in the long run.

### DIFF
--- a/src/Node.cpp
+++ b/src/Node.cpp
@@ -55,8 +55,14 @@ Node::Node(int recvLength, int prePadding, int postPadding, bool useDefaultLoop)
         nodeData->preAlloc[i] = nullptr;
     }
 
+#if OPENSSL_VERSION_AT_LEAST(1, 1)
+    // If we're compiling against OpenSSL 1.1.0, use TLS instead of SSLv23
     nodeData->clientContext = SSL_CTX_new(TLS_client_method());
     SSL_CTX_set_min_proto_version(nodeData->clientContext, TLS1_VERSION);
+#else
+    nodeData->clientContext = SSL_CTX_new(SSLv23_client_method());
+    SSL_CTX_set_options(nodeData->clientContext, SSL_OP_NO_SSLv3);
+#endif
 }
 
 void Node::run() {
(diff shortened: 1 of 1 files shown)


alexhultman:
Problem is Node.js 10 just upgraded to OpenSSL 1.1.0 so that means all versions below will fail to build.

FiniteReality:
Ah, in that case I'll add a check against OPENSSL_VERSION_NUMBER for the new code.

witaly-iwanow:
So what's exactly the purpose except breaking compatibility with 1.0.x? SSLv23_client_method is still there in 1.1.x and supports all flavours of SSL/TLS, if you don't want the former, you just block it with SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3

kapouer:
[link to openssl documentation showing SSLv23_client_method is deprecated](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_new.html#NOTES)

FiniteReality:
@witaly-iwanow This is not true. I was having issues when connecting to a TLS 1.2 websocket, and after debugging the lib I discovered it was due to OpenSSL returning that it failed to negotiate a protocol to connect with. After changing the method to TLS_client_method, everything worked correctly.

Furthermore, in OpenSSL 1.1.0, SSLv2 support was removed, and since SSLv3 is explicitly disabled as-is, I saw no need to support it which is why I set the minimum to TLS 1.0.

alexhultman:
In any case - SSL will be part of uSockets and I plan to start work on that as next step. SSL will be a completely separate layer above us_socket and us_socket_context in uSockets.

us_ssl_socket and us_ssl_socket_context

This means WebSocket<uWS::SERVER> will be WebSocket<uWS::SSL, uWS::SERVER> and same with Groups.

alexhultman:
OpenSSL 1.1 is required in next major. It makes use of TLS_server_method

WildDylan:
fatal error: 'openssl/opensslv.h' file not found

How to fix that

PR #775, opened 2018-06-12
Call shutdown() also when other direction is already closed

not-implemented:
This avoids vanishing of the SSL session cache. When calling SSL_free() on a non-shutdowned connection, the session is considered bad, and removed from the session cache:

> the SSL_SENT_SHUTDOWN flag is set and a currently open session is considered closed and good and will be kept in the session cache for further reuse.

https://www.openssl.org/docs/man1.0.2/ssl/SSL_shutdown.html

You can verify this with https://www.ssllabs.com/ssltest/index.html ... and check for:

> Session resumption (caching): Yes

### DIFF
--- a/src/HTTPSocket.cpp
+++ b/src/HTTPSocket.cpp
@@ -282,6 +282,7 @@ void HttpSocket<isServer>::onEnd(uS::Socket *s) {
         httpSocket->cancelTimeout();
     }
 
+    httpSocket->template shutdown();
     httpSocket->template closeSocket<HttpSocket<isServer>>();
 
     while (!httpSocket->messageQueue.empty()) {
--- a/src/WebSocket.cpp
+++ b/src/WebSocket.cpp
@@ -296,6 +296,7 @@ void WebSocket<isServer>::onEnd(uS::Socket *s) {
         webSocket->cancelTimeout();
     }
 
+    webSocket->template shutdown();
     webSocket->template closeSocket<WebSocket<isServer>>();
 
     while (!webSocket->messageQueue.empty()) {


alexhultman:
So this is essentially an optimization and feature?

not-implemented:
IMHO it's generally better/RFC-conform to shutdown() a socket before closing, so it is closed with TCP FIN+ACK packages. When closing a socket directly it is hard closed with a RST packet which should be used only on errors.

The problem with the SSL session cache is "just" a side effect here - the one which we noticed while testing the SSL connection ;-)

So this can also be classified as a bug.

alexhultman:
The server already does proper shutdown for non-SSL so this change breaks that (or at least is incorrect). For SSL I'm pretty sure it also is correct, I know it uses SSL_shutdown. In any case, this is obviously not a critical bug and I rather not fidget with such fundamental things this late in v0.14. Everything from this point on will be about v0.15 or v1.0 where all of this garbage code is rewritten in uSockets.

not-implemented:
Okay I see ... for the case that the server closes the connection I agree - then shutdown() is maybe called twice now - I forgot that. But for the case, the client closes the connection, no shutdown() is currently called if I am right.

But you are right it's not a critical bug. I expected a bigger traffic overhead here when session reuse does not work. But it seems the newer "SSL session tickets" have superseded the SSL sessions anyway. This patch currently runs in production for ~20k clients and we don't see a significant difference besides the succeeding SSL-Labs test ;-)

not-implemented:
No worries ;-) I just wanted to confirm that it is not a bigger SSL-Traffic problem in live environments - at least much smaller as I initially expected ;-)

Thanks!

alexhultman:
v0.14 is a complete garbage heap compared to what I'm working on now. Especially SSL will be a lot better integrated.

not-implemented:
Very cool ;-) I'm looking forward to the next version ;-)

abrownsword:
And just as I was about to start kicking the tires on v0.14 ... is there a roadmap or something for this library and uSockets?

alexhultman:
@abrownsword Tires are of good condition and has been for 2 years. This is a non-critical standards noncompliance, not a 5 GB/sec memory leak.

abrownsword:
Yeah, I got that much... but i posted here because I don’t know where is a good place to ask about your roadmap, since you’ve mentioned post-v0.14 efforts on-going. I don’t want to do a bunch of work to use a library that’s going to be put out to pasture soon.

alexhultman:
I tested the server with Wireshark and it does send proper FIN-response to a peer shutdown. It's really only an issue for SSL.

alexhultman:
What if you just add a SSL_shutdown call for this case?

not-implemented:
Ah ok, cool. Hmm, after reading some more docs, I'm still not 100% sure what's the "perfect/proper" way to shutdown a SSL/TLS-connection:

- I thought SSL_shutdown() is the SSL counterpart of shutdown() ... but it seems to be more a "high-level" shutdown that does not shutdown the socket ... and it can be called twice to receive the "close-notify" from the other side (but it's okay to call it only once)
- So, maybe is this the correct way for this use-case (?): SSL_shutdown() + shutdown(SHUT_RDWR) + close()  ??
- I think, it should not be a problem to do the same procedure if the other party already closed the connection ... some calls should just be a NOOP then (I hope).
- Of course shutdown() should not be called twice in any case (this is probably wrong in my current pull-request)

But where do you think we can just add a SSL_shutdown() call?

not-implemented:
Yes ... it's definitely easier to handle this in the rewritten uSockets implementation. No problem ;-)

One more thing regarding the socket close: It would be helpful to have a **callback in `closeSocket()`** in Socket.h - i.e. to collect some **metrics about read/written bytes** ... or to **cleanup some SSL specific things** (ECDH curves and such things). The `onDisconnection()` is too early as it is called in `Socket.close()` before the `OP_CLOSE` is sent.

Also it would be nice to count the real unencrypted bytes sent/received over a connection (with HTTP-Upgrade, WebSocket headers etc. on socket level). There is currently no way to collect the exact counters - even at application level, right? By the way for SSL connections the encrypted traffic can be already monitored with `BIO_number_read(SSL_get_rbio(sslSocket));` and `BIO_number_written(SSL_get_wbio(sslSocket));` ... when having a callback in `closeSocket()`, this could already be handled at application level. Both metrics (encrypted and unencrypted traffic) makes sense I think.

Does it make sense to open new feature-request-pull-requests for that? Or do you already plan such features? Or did I oversee something and it is already possible? Should I wait until the uSockets rewrite?

alexhultman:
Sounds like you want to implement a middle layer plugin in uSockets instead. You can do all of this in the layered design I'm working on. Every SSL byte sent in the server goes through the TCP socket so you can easily tap in to it and count.

alexhultman:
I mean, you should seriously have a look in uSockets ssl.c source code, it needs fixing.

not-implemented:
Oh great, I didn't notice that the uSockets implementation is already on GitHub in a separate repo ... I'll have a look! Thanks :-)

PR #776, opened 2018-06-24
suppress uninitialized warning

chaoran:
cb is not initialized, which will generate a warning/error at compile.

### DIFF
--- a/src/Libuv.h
+++ b/src/Libuv.h
@@ -102,6 +102,7 @@ struct Poll {
     Poll(Loop *loop, uv_os_sock_t fd) {
         uv_poll = new uv_poll_t;
         uv_poll_init_socket(loop, uv_poll, fd);
+        cb = nullptr;
     }
 
     Poll(Poll &&other) {


alexhultman:
What compiler errors because of this?

chaoran:
I am using g++-7.2.0 and set flags "-O2 -Wall -Werror -DUSE_LIBUV".

```bash
$ g++ --version
g++ (Ubuntu 7.2.0-1ubuntu1~16.04) 7.2.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
```
$ make
make `(uname -s)`
make[1]: Entering directory '/home/chaoran/src/triumph/uWebSockets-0.14.8'
g++   -std=c++11 -O2 -Wall -Werror -DUSE_LIBUV -I src -I../bazel-triumph/external/libuv/include -shared -fPIC src/Extensions.cpp src/Group.cpp src/Networking.cpp src/Hub.cpp src/Node.cpp src/WebSocket.cpp src/HTTPSocket.cpp src/Socket.cpp src/Epoll.cpp -s -o libuWS.so
In file included from src/Backend.h:9:0,
                 from src/Networking.h:69,
                 from src/WebSocketProtocol.h:5,
                 from src/WebSocket.h:4,
                 from src/Group.h:4,
                 from src/Hub.h:4,
                 from src/Hub.cpp:1:
src/Libuv.h: In member function ‘void uWS::Hub::upgrade(uv_os_sock_t, const char*, SSL*, const char*, size_t, const char*, size_t, uWS::Group<true>*)’:
src/Libuv.h:109:12: error: ‘s.uS::Poll::cb’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
         cb = other.cb;
         ~~~^~~~~~~~~~
src/Hub.cpp:243:16: note: ‘s.uS::Poll::cb’ was declared here
     uS::Socket s((uS::NodeData *) serverGroup, serverGroup->loop, fd, ssl);
                ^
cc1plus: all warnings being treated as errors
Makefile:8: recipe for target 'Linux' failed
make[1]: *** [Linux] Error 1
```

alexhultman:
So don't enable -Wall -Werror. You practically cripple your compiler intentionally and most of all software won't compiler with those flags.

chaoran:
This has nothing to do with what compiler flags I use. The compiler is complaining about a legitimate problem. And I think it should be fixed. `cb` is being used uninitialized. This will break most of (if not all) tools that checks for use of uninitialized values. You choosing not to fix it is completely illogical.

alexhultman:
This is completely incorrect and not up for discussion. You need to have a deeper look in the code and stop listening blindly to those absurd false positive warnings that you promote to full on errors. It's not debatable.

chaoran:
LOL. Oh please enlighten me. Why it’s incorrect to initialize it to nullptr?

alexhultman:
No. That's not how it works. You please enlighten me why it is broken. You refer to checkers like Valgrind and AddressSanitizer. Did you ever_actually_ run any of these? If not, please do so _before_ refusing to listen. Go ahead and do this now and give me a report of the flaw.

chaoran:
It is broken because you didn’t initialize it at construction. If user do not call `setCb`, they are screwed. Their program’s behavior is undefined. It may not even break, because the uninitialized `cb` may be actually be a different function pointer, causing the error goes undetected.

In this case, Valgrind or Address Sanitizer will surely report an error. I don’t have time to create a test case for such a simple logic error.

Initializing `cb` to nullptr ensures that their mal-formed program will surely crash if they didn’t call `setCb`. This could potentially save hours of their debugging time.

I can list a number of reasons why it is bad to left the value uninitialized. I don’t think there is any legitimate reason why leaving it uninitialized is good. If you have any reason to not merge the PR, please enumerate.

alexhultman:
> Valgrind or Address Sanitizer will surely report an error.

Okay. So show me. Do the test. Put your money where your mouth is. Give me the report.

chaoran:
I don’t need to show you. I don’t really care if you fix it or not. I created the PR as a courtesy. I think I already made my point super clear. It is sad that you don’t understand the harm of uninitialized value.

And you didn’t list even one reason why it is bad to merge the PR. This discuss is a complete waste of time.

alexhultman:
> I don’t need to show you.

Yes. Yes you do have to.

> It is broken because you didn’t initialize it at construction.

No. No it is not. Ridiculously incorrect.

> I don’t have time to create a test case for such a simple logic error.

Yet you expect *me* to have time with *you*?

> Valgrind or Address Sanitizer will surely report an error.

You talk a lot and do nothing to *actually* test if this is true or not. If you *would* do the test and do *listen* to people who knows about this, you would actually learn one or two things.

You would, of course, also understand you're wrong and that calling a nullptr is not any better than any of your made up fantasy flaws. The only thing you are solving is your delirium and placebo bias because now your compiler no longer reports this as an error. You do understand that if your claim were real, the fact that cb is nullptr does not fix anything. It would still segment fault due to accessing memory at address 0.

chaoran:
There is a difference between `always seg fault` and `undefined behavior`. If you think they are the same, there’s no need to discuss further.

chaoran:
There is really nothing to learn from the discussion, where all you say is condescending judgement without showing any actually fact.

abrownsword:
Perhaps cb ought to be initialized to a function that throws an assert or otherwise defined error behaviour.

alexhultman:
Look. Dude. I know your kind. I've met a lot of people like you. I'm a college droput for a reason.

I have tried to point you in the right direction so that you can learn but you refuse to check your claims with AddressSanitizer or Valgrind and find out first hand that you're wrong about this.

If you don't want to learn or listen to me explaining my own code then there is really no need for me to spend time on you.

Take care

chaoran:
I do want to learn. Please explain you code.

alexhultman:
You can begin by compiling a simple hello world server with asan enabled. This would immediately put an end to this back and forth inflated ego competition.

chaoran:
I think what you’re saying is that all your code path have properly initialized `cb`.

I think you can start by running asan on this code.
```
#include <uws/Libuv.h>

int main() {
	Poll poll(…);
	poll.start(…);
}
```

Of course you need to fill in the arguments. You can try it and see the result.

alexhultman:
> I think what you’re saying is that all your code path have properly initialized `cb`.

Bingo! :sparkles: :star: :tadaa:

chaoran:
That doesn’t mean your code doesn’t have bug.

Your Libuv.h is an exposed interface. And user can directly include it in their code and create a Poll object without calling `sebCb`, which happens to be how I used it in my code. If you don’t want them to do that, you should probably hide the interface from the user, or document that user should always call `setCb` when using `Poll` directly.

alexhultman:
Welcome to C++. If you don't know what you're doing, you're fucked.

Same kind of argument can be said about literally every interface ever created. Many libraries require some kind of init procedure and will simply crash if not properly initialized. OpenSSL for one.

Basically, you can't blame an interface if you're using it invalidly.

chaoran:
OpenSSL document that you should call init. You don’t document the behavior. And it is an interface that uses have access to. I have been writing C/C++ code for more than 10 years now. So I know what I’m doing.

Look, Dude. All I’m saying is that there are some benefits to initialize it to nullptr (turning an ‘undefined behavior’ to ‘always crash’; suppress compiler warning; remind your user that they use it wrong; just to name a few). And there is no downside at all in doing so. So I don’t understand your strong opinion to not merge it.

Is there any reason other than admitting your code has a bug hurts your ego?

alexhultman:
> You don’t document the behavior.

You are free to PR a documenting comment.

Using polls directly is not very common and those who use it has done so without any issues so far.

The behavior is what it is and is so for many reasons.

alexhultman:
> Is there any reason other than admitting your code has a bug hurts your ego?

This code has had tons and tons of bugs. This is not one of them.

chaoran:
Could you please list one of the reasons that you insists it being uninitialized at construction? If it is because performance reasons, I assure you that the nullptr assignment will be optimize away by compiler.

If you can enumerate a reason for keeping the behavior as it is, I might actually learn something. Because it would be the first example I met that shows uninitialized values might be good.

alexhultman:
Mainly because I don't change things for no reason just because someone has a strong opinion about it. I am busy with next major version which will be a complete rewrite so I don't plan to merge any changes not fixing anything critical.

chaoran:
Hey, I understand people don't want to merge PR for small things when there's is a big rewrite going to happen.

But it is unfortunate that we could have avoided this useless discussion if you say that in the beginning OR didn't judge my PR as "This is completely incorrect and not up for discussion".

alexhultman:
Yes or if you listened.

abrownsword:
I see that this PR was closed, and my last comment on it ignored.  Perhaps I didn’t explain my suggestion well enough.  It will likely still be relevant in the future...

Rather than initializing a callback to nullptr or leaving it uninitialized, I suggest that it be initialized to a stub function that asserts and/or produces a useful message for the user of the library.  That way they won’t have to deal with a crash coming out of the library but will instead have a deterministic and informative failure.  The cost in performance should be negligible for a noteable gain in usability and one less annoying compiler warning.

PR #777, opened 2018-06-28
CMake build support

mmusial:
Support for building and installing with CMake.
By default static version of library is built (Fix for Windows build in pull-request).
Set CMake build-in variable BUILD_SHARED_LIBS=ON if shared library is needed.


alexhultman:
Cmake = piss in my face

PR #778, opened 2018-07-04
Allow to set send/recv buff size

noobxinyu:


alexhultman:
Aren't there settable via config files?

noobxinyu:
It would be a bit intrusive to the system and not flexible per application. Also a feed handler / trading gateway should have totally different settings.

alexhultman:
Maybe better to use the getFd() function and do whatever

noobxinyu:
I thought this might be parallel with cord/nodelay settings. But anyway it is your call.

PR #779, opened 2018-07-22
Moved UV_READABLE, UV_WRITABLE, and uv_os_sock_t into uS namespace

izaid:
At present, the declarations of UV_READABLE, UV_WRITABLE, and uv_os_sock_t make it not possible to include both `uv.h` and uWS files like `Hub.h` together in the same source file (if one is using the epoll or asio backend) as these definitions conflict.

This PR Is the most minimal set of changes to remove this conflict. Perhaps it would be better to rename as follows: `UV_READABLE` -> `uS::READABLE`, `UV_WRITABLE` -> `uS::WRITABLE`, and `uv_sock_t` -> `us_os_sock_t`, which I would also be happy to do.

### DIFF
--- a/src/Asio.h
+++ b/src/Asio.h
@@ -3,11 +3,11 @@
 
 #include <boost/asio.hpp>
 
+namespace uS {
+
 typedef boost::asio::ip::tcp::socket::native_handle_type uv_os_sock_t;
 static const int UV_READABLE = 1;
 static const int UV_WRITABLE = 2;
-
-namespace uS {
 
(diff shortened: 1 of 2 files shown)


alexhultman:
This would be a nice thing to fix if 0.14 would stay for longer but all focus is currently on 0.15 and it does not have the same issue of leaking names outside of its modules like 0.14 do.

izaid:
I don't mind making the changes to 0.14 if you don't mind merging them. If you'd rather wait for 0.15 though, then I'm happy to just close the PR.

PR #780, opened 2018-08-15
Add run_one method to Libuv and Asio

Tai7sy:
Add run_one method to Libuv, Asio.
Epoll not

### DIFF
--- a/src/Asio.h
+++ b/src/Asio.h
@@ -23,6 +23,10 @@ struct Loop : boost::asio::io_service {
         boost::asio::io_service::run();
     }
 
+    void run_one() {
+        boost::asio::io_service::run_one();
+    }
+
     void poll() {
         boost::asio::io_service::poll();
     }
--- a/src/Epoll.h
+++ b/src/Epoll.h
@@ -61,6 +61,8 @@ struct Loop {
     void doEpoll(int epollTimeout);
 
     void run();
+	
+	void run_one();
 
     void poll();
 
--- a/src/Epoll.h
+++ b/src/Epoll.h
@@ -62,7 +62,7 @@ struct Loop {
 
     void run();
 	
-	void run_one();
+    void run_one();
 
     void poll();
 
--- a/src/Hub.h
+++ b/src/Hub.h
@@ -73,6 +73,7 @@ public:
     }
 
     using uS::Node::run;
+    using uS::Node::run_one;
     using uS::Node::poll;
     using uS::Node::getLoop;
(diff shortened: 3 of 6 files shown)


alexhultman:
run_one is not the naming convention used. runOne would be, but I'm wondering why you would need this? A proper solution would only need run, nothing else. Integrations should be made at the polling (eventing) layer.

Tai7sy:
I start many servers and stop some when receive some message. But there is no stop method(and it maybe hard to force terminate a server in another thread safely). 
In the beginning, I use while/poll/Sleep so i can break the whileloop when needed, then I found a run_one would be better, because it use the eventloop in libuv(or some other) rather than a stupid Sleep.
What's more, I think these i/o libraries support runOne, so there must be some situation we can use it.

alexhultman:
You need to go back to the drawing board and redesign your server how it's supposed to work instead of adding hacks to work around your own design limitations.

abrownsword:
I think the correct paradigm on libuv is to call loop->stop().  Or I guess that’s uv_loop_stop() if I’m not using a c++ wrapper.  There are also alternate run modes for loops, but I don’t like those much.  And runOne seems like a poor paradigm, particularly if your objective is to start and stop a server on demand.  If you want to act per iteration use a per iteration event, if you want to stop the loop... use stop.

alexhultman:
This lib is async -> it runs on events only. Just like Node.js the loop will automatically "fall through" once there is no queued up async events. This is how you close it down if you really want that. You can look in the tests source, there are about 50 examples of how to properly do this. If you need more info you can take a quick look at Node.js design principles and apply them here.

alexhultman:
Also note that your design will leak memory and file descriptors if you just hack in a runOnce method and exit that way. You need to let the loop run and fall through. Hub::run(); will return "fall through" once it is ready to do so.

abrownsword:
Tai7sy:  the libuv paradigm for stopping a run loop in another thread is probably to use uv_async_t, and have the handler of that event (which is on the thread of the run loop) call uv_loop_stop once it has properly processed any cleanup that needs to be done.  Trying to take control of the loop’s iteration process by calling it one iteration at a time is going to be fraught with problems, as Alex points out.

Tai7sy:
Thanks for these advice.
Now I have understood how the uWS works and the problem in my design of my program.
Just call close() and waiting for the run() to 'fall through' automatically seems to be more elegant and the right way.
Examples in the tests helped me a lot to understand and redesign the procedure of my program. Thanks again.

PR #781, opened 2018-08-21
Epoll Bug Fixes - Timer Resolution Accuracy Improved

ptaulborg:
Fixed two major bugs with Epoll timers.

1) Rewrote the timer check in Loop::run to reset the next epoll_wait() epollTimeout/delay if there was socket activity but no executed timer. This fixes timer resolution to be very accurate, regardless of i/o activity.

2) Critical bug fix where if while executing a timer callback, the callback stop()'d a different timer, the loop code would "continue" thinking the current timer was stopped. This was because it uses a Loop:: boolean that is global to that context. Added checking to make sure the current processing timer was stopped. The result of this bug would be that the timer would execute a second time immediately.

One other note is that if during the timer check loop your callback triggers the creation of a new Timer that wants to execute immediately (0ms), this entire chain will break, since it relies on a presumed index of [0] after the callback executes. This doesn't matter for my particular usage, because I'd never use timers in this way, but for sake of being super robust, just making full disclosure. That particular usage bug exists in both mine and the original branch.

### DIFF
--- a/src/Epoll.cpp
+++ b/src/Epoll.cpp
@@ -35,20 +35,28 @@ void Loop::doEpoll(int epollTimeout) {
         callbacks[poll->state.cbIndex](poll, status, readyEvents[i].events);
     }
 
-    while (timers.size() && timers[0].timepoint < timepoint) {
-        Timer *timer = timers[0].timer;
-        cancelledLastTimer = false;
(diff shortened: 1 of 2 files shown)


alexhultman:
Alright seems like you did some research here, trusting you this works

PR #826, opened 2019-02-01
-

guffy1234:

### DIFF
--- a/src/HttpRouter.h
+++ b/src/HttpRouter.h
@@ -37,7 +37,7 @@ namespace uWS {
 template <class USERDATA>
 class HttpRouter {
 private:
-    static const unsigned int MAX_URL_SEGMENTS = 100;
+    static constexpr unsigned int MAX_URL_SEGMENTS = 100;
 
     /* Basically a pre-allocated stack */
     struct RouteParameters {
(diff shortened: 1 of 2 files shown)


src/HttpRouter.h
```diff
@@ -257,27 +221,21 @@ class HttpRouter {
...
-
-        /* We have that method on record, let's iterate it */
-        if (parent != &tree) {
-            return executeHandlers(parent, 0, userData);
-        } else {
-            /* We did not find any handler for this method.
-             * You may want to re-route with "*" as method. */
-            return false;
-        }
+		for (auto &p : tree.children) {
```
alexhultman: This is basically

```
for (auto &c : tree->children) {
	if (c->name == method) {
		return executeHandlers(c.get(), 0, userData);
	}
}
return false;
```

guffy1234:
Yes. Have no idea why github shows changes this way.
This is the file
https://github.com/guffy1234/uWebSockets/blob/master/src/HttpRouter.h

src/HttpRouter.h
guffy1234: Yes. Have no idea why github shows changes this way.
This is the file
https://github.com/guffy1234/uWebSockets/blob/master/src/HttpRouter.h

alexhultman:
Hello,

I can look at this and see what's easy to put in. The router is still WIP and should be rewritten completely when it's behavior is known.

guffy1234:
Ok. Honestly it looks like designed to support high amount of routes. You trying to save space in the nodes with indirect handlers. May be I'm not too familiar with web applications, but imho probably it's unusual case. And tree-like routing makes difficult to introduce regex matches, especially multi segment patterns

alexhultman:
It doesn't do regex. Regex are slow very slow

guffy1234:
I have changed my opinion. Tree based routing is enough. Whenever i will need something custom i can do it as second layer routing in a handler. Like filesystem streamer does.

PR #847, opened 2019-03-05
Add method to get all headers on the request

kyle-mccarthy:
I would like to be able to get all the headers for a given request, with the intention of being able to add this into the uWebSockets.js repo. [Commit for uWebSocket.js on fork](https://github.com/kyle-mccarthy/uWebSockets.js/commit/feee90b78a5faecbc72c3d098834e4d999a29278).

I want to apologize in advance if I am doing something wrong or a bad way, I don't actually know c++.

### DIFF
--- a/src/HttpParser.h
+++ b/src/HttpParser.h
@@ -62,6 +62,16 @@ public:
         return std::string_view(nullptr, 0);
     }
 
+    std::map<std::string_view, std::string_view> getHeaders() {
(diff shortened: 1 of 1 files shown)


alexhultman:
This is not how we solve problems here. Algorithmically this is very poor. You want a simple loop.

alexhultman:
Unless you're doing a proxy, one should never loop over all headers. You're solving a problem that isn't there

Issue #851, opened 2019-03-13
subprotocol


alexhultman:

alexhultman:
The first websocket subprotocol should be sent back, separated by comma in sec-websocket-subprotocol

alexhultman:
This should be a 1 line fix

alexhultman:
Fixed

PR #856, opened 2019-03-18
Add 'Sec-Websocket-Protocol' support

mtt0:
#851 Add 'Sec-Websocket-Protocol' support

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -90,6 +90,7 @@ public:
         CompressOptions compression = DISABLED;
         int maxPayloadLength = 16 * 1024;
         int idleTimeout = 120;
+        SecWebSocketProtocol supportSubprotocol;
         fu2::unique_function<void(uWS::WebSocket<SSL, true> *, HttpRequest *)> open = nullptr;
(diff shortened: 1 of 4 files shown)


src/App.h
```diff
@@ -149,6 +150,14 @@ struct TemplatedApp {
                     ->writeHeader("Connection", "Upgrade")
                     ->writeHeader("Sec-WebSocket-Accept", secWebSocketAccept);
 
+                std::string_view secWebSocketProtocol = req->getHeader("sec-websocket-protocol");
+                if (secWebSocketProtocol.length()) {
+                    auto ret = behavior.supportSubprotocol.selectBest(secWebSocketProtocol);
```
alexhultman: I think selectBest could simply select the first protocol in the list passed to us. Basically take sec-websocket-protocol and just trim it by first comma.

mtt0: Yes. But I think **a websocket server may provides some subprotocols for communicate with clients, only the clients who support one of these subprotocols asked by server will be accept.** Thus, the server should choose the first protocol in the list which is also existed in server's protocol list. Am I think too much or just totally wrong.... not  sure about that.

mtt0:
Simply select the first protocol in the list passed to us.

PR #857, opened 2019-03-18
fix compile issues

johnli1210:
1. Fix wrong format of brace-enclosed initialization of WebSocketBehavior in examples
2. In AsyncSocket some member methods are declared as protected, but later call by object


mtt0:
It is fine under clang++ while compile error under g++ @johnli1210

PR #869, opened 2019-04-02
Store HTTP response data before open event

### LINKED ISSUE #867: Cannot remove fallback buffer of HttpParser before emitting websocket open event
I got 2 similar segmentation faults that are most likely caused by malformed requests. Here is the log output:
```
2019-03-23T03:08:14.936Z wt-tracker:uws-tracker-requests 0.0.0.0 443 ws-open url: / query:  origin: https://www.streamlive.to total: 286
2019-03-23T03:08:29.554Z wt-tracker:uws-tracker-requests 0.0.0.0 443 ws-open url: ^@ query:  origin: https://www.streamlive.to total: 283
Segmentation fault (core dumped)
```
First line is how looks normal request. Second line is how looks fault request. Symbol ^@ from the log refers to 0x00.

uWebSockets.js v15.8.0

This is how I gather the logs:
https://github.com/Novage/wt-tracker/blob/master/lib/uws-tracker.ts#L91

mrlika:
Cannot remove fallback buffer of HttpParser before emitting WebSocket open event
Fixes #867

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -189,11 +189,8 @@ public:
                 /* This will add our mark */
                 res->upgrade();
 
-                /* Move any backpressure */
-                std::string backpressure(std::move(((AsyncSocketData<SSL> *) res->getHttpResponseData())->buffer));
-
-                /* Destroy HttpResponseData */
-                res->getHttpResponseData()->~HttpResponseData();
+                /* Temporary store HTTP response data that may be referenced by req object */
+                HttpResponseData<SSL> httpResponseData(std::move(*res->getHttpResponseData()));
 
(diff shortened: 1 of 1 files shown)


alexhultman:
This is almost identical to the solution I was going for BUT there is one issue - moving the std::string does not guarantee it's .data() remains the same. Small strings are (potentially) stored on the stack.

I'm thinking more like changing from std string to raw pointer of size FALLBACK_EXPECTED of about 2kb or so.

mrlika:
You are right. SSO breaks this. As an option std::string::reserve may be called to reserve 24 bytes or more to avoid SSO: https://github.com/elliotgoodrich/SSO-23#comparison

PR #906, opened 2019-06-13
Clarify getHeader argument meaning

zxey:
I tried to use getHeader method on HttpRequest and found out that it did not perform case insensitive search on headers.

It is specified by [RFC 2616 - "Hypertext Transfer Protocol -- HTTP/1.1"](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) that headers should be case insensitive:
> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

### DIFF
--- a/src/HttpParser.h
+++ b/src/HttpParser.h
@@ -86,9 +86,9 @@ public:
         didYield = yield;
     }
 
-    std::string_view getHeader(std::string_view header) {
+    std::string_view getHeader(std::string_view lowerCasedHeader) {
         for (Header *h = headers; (++h)->key.length(); ) {
-            if (h->key.length() == header.length() && !strncmp(h->key.data(), header.data(), header.length())) {
+            if (h->key.length() == lowerCasedHeader.length() && !strncmp(h->key.data(), lowerCasedHeader.data(), lowerCasedHeader.length())) {
                 return h->value;
             }
         }


src/Utilities.h
```diff
@@ -60,6 +60,35 @@ inline int u32toa(uint32_t value, char *dst) {
     return ret;
 }
 
+inline bool asciiEqualsCaseI(const char * str1, const char * str2, size_t num)
```
dvergeylen: This could probably be more concise e.g:

```cpp
bool iequals(const string& a, const string& b)
{
    return std::equal(a.begin(), a.end(),
                      b.begin(), b.end(),
                      [](char a, char b) {
                          return tolower(a) == tolower(b);
                      });
}
```
(unashamedly taken from https://stackoverflow.com/a/4119881)

alexhultman:
All header keys are lower cased, so you just have to specify the key in lower case.

alexhultman:
https://github.com/uNetworking/uWebSockets/blob/master/src/HttpParser.h#L89

If anything, the argument can be named lowerCasedHeader

zxey:
> All header keys are lower cased, so you just have to specify the key in lower case.

Oh, nice, I just changed the argument to be named `lowerCasedHeader` then.

ddevienne:
lowerCase**d**Header implies it might be lower-cased by the implementation of that function.
While lowerCaseHeader header implies that argument is expected to be lower-case IMHO.
I'm not a native speaker, but that how it seems to me. FWIW.

alexhultman:
headerInLowerCase

Issue #805, opened 2019-01-02
Missing way to read HttpRequest body/data for POST/PUT...


christoth:
Apologies if I missed it, but I can't find a way to read the request body (the data after the headers) in HttpRequest. 

This would be needed in handling POST/PUT/... requests.

Thanks in advance

alexhultman:
res->read([](std::string_view data) {

});

alexhultman:
No response closing

christoth:
Thanks for pointing me in the right direction.

Is there a way to tell when the request data is finished? e.g. zero length data

I'm currently reading the content-length header, appending data to by own buffer and comparing the length of the buffer to content-length.

Are you open to me adding an echo HTTP example (using POST and returning the request body) either on it's own or to the echo server example?
 
Thanks again

alexhultman:
I added a bool "last" and renamed it to onData. I've tested this on Windows and it seems to work. Probably should add more helper functions later on.

Thanks for reporting

```c++
.post("/file", [](auto *res, auto *req) {
        res->onData([res](std::string_view data, bool last) {
            std::cout << "Got data: " << data.length() << ", it's the last: " << last << std::endl;

            if (last) {
                us_listen_socket_close(listen_socket);
                res->end("Thanks for the data!");
            }
        });
```

christoth:
Using onData works with single requests.

When I test with concurrent requests it times-out, but it doesn't crash now, which is great and it will respond to single requests after the short loadtest - previously with read() it just crashed. 

It looks like concurrent requests trigger onData(), but last is never true?

Here's my code:

        }).post("/echo", [](auto *res, auto *req) {
          const uint64_t contentLength = std::stoi(std::string(req->getHeader("content-length")));
          std::cout << ">: " << contentLength << " ";
          if(contentLength == 0) {
            res->end("'/echo' request needs a payload");
            std::cout << "contentLength == 0" << std::endl;
            return;
          }

          std::string requestData;

          res->onData([&](std::string_view data, bool last) {
            if (last) {
              if (requestData.size() == 0) {
                res->end(data);
                std::cout << "<: " << data.length() << std::endl; 
              } else {
                requestData.append(data.data(), data.size());
                res->end(requestData);
                std::cout << "[" << data.size() << "] <: " << requestData.length() << std::endl;
              }
              return;
            } else {
              if (requestData.length() == 0) {
                requestData.reserve(contentLength);
              }
              requestData.append(data.data(), data.size());
[...]

PR #907, opened 2019-06-14
allow delayed response onData handling

sigidagi:
@alexhultman  following your last POST request example in [issues/805](https://github.com/uNetworking/uWebSockets/issues/805)

 `res->end("Thanks for the data!");` inside onData lambda will terminate with 

> Error: Returning from a request handler without responding or attaching an abort handler is forbidden!

### DIFF
--- a/src/HttpContext.h
+++ b/src/HttpContext.h
@@ -168,7 +168,7 @@ private:
                 }
 
                 /* Returning from a request handler without responding or attaching an onAborted handler is ill-use */
-                if (!((HttpResponse<SSL> *) s)->hasResponded() && !httpResponseData->onAborted) {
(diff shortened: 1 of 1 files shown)


alexhultman:
You need to attach an abort handler, call onAborted. That is the only restriction

sigidagi:
What is rationale of having extra handler _"onAborted_"  when handling POST/PUT request using _"onData"_?

alexhultman:
When you get a request you either immediately respond or attach various callbacks, one of which is the required onAborted. Abort is basically when the connection is broken which can happen any time from request handler until fully responded

xingyun86:
> What is rationale of having extra handler _"onAborted_" when handling POST/PUT request using _"onData"_?

Can you  tell me how to attach  an abort handler? Where to do this?

xingyun86:
I had get it. Thanks you very much!

Issue #920, opened 2019-08-19
Compiled failed on centos 7.6 with g++ 7.3.1 


ForrestSu:
GCC:  g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)
platform:  CentOS Linux release 7.6.1810 (Core)
I added a header file  `climits` , fix this compile error.

```
src/WebSocketExtensions.h: In member function ‘int uWS::ExtensionsParser::getToken(const char*&, const char*)’:
src/WebSocketExtensions.h:59:23: error: ‘SHRT_MIN’ was not declared in this scope
         static_assert(SHRT_MIN > INT_MIN, "Integer overflow fix is invalid for this platform, report this as a bug!");
                       ^~~~~~~~
src/WebSocketExtensions.h:59:23: note: suggested alternative: ‘WINT_MIN’
         static_assert(SHRT_MIN > INT_MIN, "Integer overflow fix is invalid for this platform, report this as a bug!");
```

PR #921, opened 2019-08-19
Fix compile error: 'INT_MIN' was not declared in this scope

ForrestSu:
for fix issue #920 , and add make clean


alexhultman:
Could you separate them as two PRs?

ForrestSu:
> Could you separate them as two PRs?

ok👌

Issue #926, opened 2019-08-21
Expose zlib options for DEDICATED_COMPRESSOR


not-implemented:
To optimize memory usage when using DEDICATED_COMPRESSOR it would be nice to make the hardcoded values for `deflateInit2()` in https://github.com/uNetworking/uWebSockets/blob/master/src/PerMessageDeflate.h#L69 configurable.

Is it worthwhile to write a pull-request for that? I can try to add the options to WebSocketBehavior struct?

Example: Currently we use `windowBits = -11, memLevel = 3` with almost the same compression ratio for simple JSON messages ... but reduced zlib RAM usage to ~12-16 KB / client, instead of 300 KB (important when having ~25.000 connections ;-))

alexhultman:
Yes that could make sense but I guess you need to also update the negotiation of permessage-deflate options

alexhultman:
I would like to have simplified options:

From no memory overhead to lots of memory overhead and then many steps in between. I rather have simplified options people can easily grasp than strange knobs few understand.

Maybe simply have an integer going from 0 (off), 1 (shared), 2 (low memory usage), 3 (lots of memory usage), and so on

alexhultman-2fa:
There are formulas for calculating those settings to reach a certain memory usage:

2kb
4kb
8kb
16kb

And so on. It could be made as simple as a set of flags you can specify like COMPRESSOR_16KB,

Anding with DEDICATED_COMPRESSOR.

That way it can be backwards compatible and impossible to specify wrong settings and it becomes easy to tinker with settings in regards to memory usage

alexhultman:
Like I wrote in the PR, I have added DEDICATED_COMPRESSOR_2KB and all the way up to DEDICATED_COMPRESSOR_256KB so that it is super easy to pick one based on ones memory budget. Thanks

not-implemented:
@alexhultman: Very cool, thanks! I hope I will have some time in the near future to test/integrate this (and your changes from december - the new cork() and the improved router) :+1:

not-implemented:
@alexhultman: Works well, thanks!

PR #928, opened 2019-08-26
Implemented CompressorOptions

not-implemented:
I implemented the CompressorOptions (#926) now for the dedicated-compressor (not for the shared one). Please check if it fits your requirements.

By the way: The negotiation of permessage-deflate options (like max-window-bits) does not need to be changed here - as theese are just hints (which are currently ignored anyway). The actual negotiation of deflate-options is done inside the zlib-protocol. (At least in 0.14 we patched just the deflateInit2() params and it runs fine in production :-))

I also renamed CompressOptions -> CompressorType in a separate commit - to avoid naming confusion.
But it's an incompatible change maybe - it's just a proposal :-)

By the way: I added the new option also to the examples - as in C++<=17 all members have to be initialized. What's your plan here for the future? Just wait for C++20, where you can initialize specific members only? :-)

### DIFF
--- a/examples/BroadcastingEchoServer.cpp
+++ b/examples/BroadcastingEchoServer.cpp
@@ -10,6 +10,7 @@ int main() {
     uWS::App().ws<PerSocketData>("/*", {
         /* Settings */
         .compression = uWS::SHARED_COMPRESSOR,
+        .compressorOptions = {},
         .maxPayloadLength = 16 * 1024,
         .idleTimeout = 10,
         /* Handlers */
--- a/examples/EchoServer.cpp
+++ b/examples/EchoServer.cpp
@@ -10,6 +10,7 @@ int main() {
     uWS::App().ws<PerSocketData>("/*", {
         /* Settings */
         .compression = uWS::SHARED_COMPRESSOR,
+        .compressorOptions = {},
         .maxPayloadLength = 16 * 1024,
 	.idleTimeout = 10,
         /* Handlers */
--- a/examples/EchoServerThreaded.cpp
+++ b/examples/EchoServerThreaded.cpp
@@ -18,6 +18,7 @@ int main() {
             uWS::App().ws<PerSocketData>("/*", {
                 /* Settings */
                 .compression = uWS::SHARED_COMPRESSOR,
+                .compressorOptions = {},
                 .maxPayloadLength = 16 * 1024,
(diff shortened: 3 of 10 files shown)


alexhultman:
Interesting

alexhultman:
> The actual negotiation of deflate-options is done inside the zlib-protocol

This is interesting, needs to be tested in Autobahn but seems like it could make sense

examples/BroadcastingEchoServer.cpp
```diff
@@ -10,6 +10,7 @@ int main() {
     uWS::App().ws<PerSocketData>("/*", {
         /* Settings */
         .compression = uWS::SHARED_COMPRESSOR,
+        .compressorOptions = {},
```
alexhultman: This could be dedicatedCompressorOptions, it only applies to dedicated compressors

not-implemented: Yes you are right - it's a bit long - but I don't have a better descriptive name.

Should I change it?

alexhultman:
I have added DEDICATED_COMPRESSOR_2KB and up to 256KB options now. It is possible to compress with a smaller window than what you negotiated because the client still treats the connection as having the big window, yet we, the compressor, does not use it. So it works, but could be improved by telling the client so that it does not have to keep this big window. Thanks for reporting this improvement!

RDP1974:
Hello,

should I suggest to take a look to optimized Zlib libraries for server scalability:
http://www.htslib.org/benchmarks/zlib.html

Actually the Intel version is still the fastest in zcFastest mode (-2 level introduced in the lib).

Kind regards.

Roberto Della Pasqua
www.dellapasqua.com<http://www.dellapasqua.com>

Da: Alex Hultman <notifications@github.com>
Inviato: sabato 29 febbraio 2020 04:09
A: uNetworking/uWebSockets <uWebSockets@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Oggetto: Re: [uNetworking/uWebSockets] Implemented CompressorOptions (#928)


I have added DEDICATED_COMPRESSOR_2KB and up to 256KB options now. It is possible to compress with a smaller window than what you negotiated because the client still treats the connection as having the big window, yet we, the compressor, does not use it. So it works, but could be improved by telling the client so that it does not have to keep this big window. Thanks for reporting this improvement!

Issue #860, opened 2019-03-19
Wildcard routing ws with get


aadityataparia:
After fixing https://github.com/uNetworking/uWebSockets.js/issues/98, I am having a different problem

I am doing 

```
uWs.App()
  .ws('/*', options)
  .get('/ok', (res) => res.end('ok'))
  .get('/*', (res) => res.end('not ok'))
```

But when I go to `/ok` I get `not ok`, if I remove `.get('/*')` it works, if I remove `.ws(/*)` it works, but together they do not. 🤔

alexhultman:
This is by design - bad design but still. .ws and .get are the same in many ways so you want to separate them by URL as far as possible. The docs talk about routing a bit more.

szmarczak:
What about faking the HTTP method? Something like:

[HttpContext.h](https://github.com/uNetworking/uWebSockets/blob/master/src/HttpContext.h#L144-L152)
```diff
                typename HttpContextData<SSL>::RouterData routerData = {(HttpResponse<SSL> *) s, httpRequest};
+
+				std::string_view method = httpRequest->getMethod();
+				if (req->getHeader("sec-websocket-key").length() == 24) {
+					method = "ws";
+				}
+
-               if (!httpContextData->router.route(httpRequest->getMethod(), httpRequest->getUrl(), routerData)) {
+               if (!httpContextData->router.route(method, httpRequest->getUrl(), routerData)) {
                    /* If first pass failed, we try and match by "any" method */
                    if (!httpContextData->router.route("*", httpRequest->getUrl(), routerData)) {
                        /* If second pass fail, we have to force close this socket as we have no handler for it */
                        us_new_socket_close(SSL, (us_new_socket_t *) s);
                        return nullptr;
                    }
                }
```

[App.h](https://github.com/uNetworking/uWebSockets/blob/93b6971bf75e6360cfdc659d3a5ef5ea95b8e305/src/App.h#L139-L142)
```diff
-       return std::move(get(pattern, [webSocketContext, this, behavior = std::move(behavior)](auto *res, auto *req) mutable {
+       return std::move(ws(pattern, [webSocketContext, this, behavior = std::move(behavior)](auto *res, auto *req) mutable {
            /* If we have this header set, it's a websocket */
            std::string_view secWebSocketKey = req->getHeader("sec-websocket-key");
-           if (secWebSocketKey.length() == 24) {
				...
-			}
	...
[...]

PR #929, opened 2019-08-26
Allow different handlers for GET and WS for same route (#860)

not-implemented:
I made a pull-request out of the proposal from @szmarczak (slightly simplified).

Seems to work. Currently I do not see any drawbacks.

Btw.: I didn't fully understand the `req->setYield()` concept of the router, as it would make sense here, when multiple handlers for the same route could be registered.

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -135,7 +135,7 @@ public:
         webSocketContext->getExt()->maxPayloadLength = behavior.maxPayloadLength;
         webSocketContext->getExt()->idleTimeout = behavior.idleTimeout;
 
-        return std::move(get(pattern, [webSocketContext, httpContext = this->httpContext, behavior = std::move(behavior)](auto *res, auto *req) mutable {
+        httpContext->onHttp("ws", pattern, [webSocketContext, httpContext = this->httpContext, behavior = std::move(behavior)](auto *res, auto *req) mutable {
 
             /* If we have this header set, it's a websocket */
             std::string_view secWebSocketKey = req->getHeader("sec-websocket-key");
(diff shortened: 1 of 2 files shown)


alexhultman:
Maybe I'm a total buzzkill now but the problem is that you add more work to be done for everyone, even if you don't have the same route for Http and WebSockets.

A proper solution would change the router so that two routes can be on the same path and that it only checks for sec-websocket-key if that very route has two registered handlers - because it is only in that case it can be both.

not-implemented:
Hmm ... do you mean CPU-work or programmers-work?

CPU-work: Okay, the header lookup in HttpContext.h is done everytime then, even if no "ws" route is registered, I see. The second lookup in App.h is a little overhead, yes.
programmers-work: Should be the same - the interface does not change, or am I wrong?

But yes, maybe it is more flexible to have a list of handlers in `HttpRouter::Node` ... maybe I'll try around with `std::vector<short> handlers;` there.

Btw. (little off-topic): This would also not solve the problem to set additional custom headers for a WebSocket-Upgrade-Response - first I thought, this handler-list will be a handy interface for that - but registering a GET-handler, calling `res->setHeader()` and `req->setYield(true)` to forward it to the WS-Handler will not work - but this will be a different feature, I'll think about it :-)

szmarczak:
@not-implemented The CPU one.

I think it's needed to create a `HttpRouter::find` function and make `executeHandlers` use that `find` function. Then, when attaching a GET handler we check if there's one already. If so, move it and attach a new one:

```cpp
if (req->getHeader("sec-websocket-key").length() == 24) {
    movedHandler(res, req);
} else {
    newHandler(res, req);
}
```

The same with the WS handler, just with the inversed `if` statement.

alexhultman:
Yes that's a good solution I think - you only pay the extra cost if you really need it and only for that route

alexhultman:
Have a try with latest alpha

PR #937, opened 2019-09-19
dst size should match whats being copied into it

ink-su:
Max header value is 14 within `formatMessage`. 
Seeing as dst matches `message.length()` with a header prepended we can make dst as small or large as we want.

Changed this because I got errors when testing large messages

### DIFF
--- a/src/WebSocket.h
+++ b/src/WebSocket.h
@@ -150,7 +150,7 @@ public:
         );
 
         /* We frame the message right here and only pass raw bytes to the pub/subber */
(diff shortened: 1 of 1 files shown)


alexhultman:
You're right there are bugs in pub/sub, it is very broken currently. This PR does not really solve the issue though, as you trade one bug for a stack overflow.

ink-su:
As an alternative it's also possible to just generate the header, then append both the header and the message to the node in the publish function.

PR #944, opened 2019-10-05
Adding example for batch processing

ibsusu:
I wanted to add an example for batch processing, the problem is that in order to do it kinda-sorta-nicely I added two methods on the HttpResponse struct, a variable to the HttpResponseData struct and added a struct called AsyncProcData to the uWS namespace.  They don't belong there, I know.  I was hoping for suggestions on how I can refactor it so there is a simpler example for async processing than the AsyncFileStreamer.h.  I'm not the best programmer so it took me a couple of days to understand everything and code this up.

Thanks for creating such a great library.  This thing is sick.


ibsusu:
this code works but is bad, don't use it.  I can't delete this pr.

PR #951, opened 2019-10-17
Type casting fix

yarreg:
If we want to utilizate  uint64 type of ```intersection``` variable. We should convert 1 to 64 bit value.
Otherwise, if value of ```i``` exceeds 32 we can get an unobvious value.


alexhultman:
Could you do it (uint64_t) 1 instead, to be size explicit

yarreg:
Done

PR #971, opened 2019-12-10
Various warning fixes for windows

englercj:
I compile with Warning Level 4 (/W4) and Warnings as Errors (/WX) on windows. This may not be all the warnings, but it addresses the ones I've encountered so far. Most of them are just signed/unsigned mismatches between `size_t` and `int`.

### DIFF
--- a/src/AsyncSocket.h
+++ b/src/AsyncSocket.h
@@ -82,7 +82,7 @@ protected:
         LoopData *loopData = getLoopData();
         if (loopData->corkedSocket == this && loopData->corkOffset + size < LoopData::CORK_BUFFER_SIZE) {
             char *sendBuffer = loopData->corkBuffer + loopData->corkOffset;
-            loopData->corkOffset += size;
+            loopData->corkOffset += int(size);
             return {sendBuffer, false};
         } else {
             /* Slow path for now, we want to always be corked if possible */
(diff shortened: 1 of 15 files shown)


alexhultman:
This PR is big with many sensitive changes. I've had people try and silence warnings before only to introduce security flaws and bugs.

This is going to take some time to review, and probably piece by piece.

englercj:
No problem, I'm just hacking in a side project so there is no urgency. I wasn't really being careful or thoughtful about the changes either, so I'm sure there are adjustments that need to be made. Hoped it would be a good starting point.

alexhultman:
I have fixed a few warnings now. Do you have a lot of them remaining?

alexhultman-2fa:
I'm closing this, I have fixed all /W3 warnings now

englercj:
I do have a lot remaining still because I compile with /W4. Since this library is header-only my only options are:

1. Wrap each header I wish to use in pragmas to disable warnings.
2. Change my project to use /W3, a different warning level than I want to.
3. Use my fork where the warnings are fixed.

alexhultman-2fa:
Yeah there are more to fix but many of them got fixed recently.

"Fixed" is of course nonsense as there is no difference between a "fixed" warning and a non-fixed one if the fix is just adding whatever implicit type conversion as an explicit one.

I'll continue to remove warnings with time though.

PR #998, opened 2020-01-26
Optionally prevent sending of "uWebSockets" header

dmitigr:
This could be done by defining UWS_HTTPRESPONSE_NO_WRITEMARK macro.

### DIFF
--- a/src/HttpResponse.h
+++ b/src/HttpResponse.h
@@ -75,10 +75,12 @@ private:
         httpResponseData->state &= ~HttpResponseData<SSL>::HTTP_RESPONSE_PENDING;
     }
 
+#ifndef UWS_HTTPRESPONSE_NO_WRITEMARK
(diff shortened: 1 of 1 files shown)


src/HttpResponse.h
```diff
@@ -75,10 +75,12 @@ struct HttpResponse : public AsyncSocket<SSL> {
         httpResponseData->state &= ~HttpResponseData<SSL>::HTTP_RESPONSE_PENDING;
     }
 
+#ifndef UWS_HTTPRESPONSE_NO_WRITEMARK
     /* Called only once per request */
     void writeMark() {
         writeHeader("uWebSockets", "v0.17");
```
alexhultman-2fa: You can actually just move the macro around this one line. No need to put the macro at multiple places.

dmitigr: Done.

PR #1042, opened 2020-04-28
standards compliant installation scheme for Makefile

epsilon-0:
added necessary paths and install commands to
works on both linux and BSD

Tested on OpenBSD and Gentoo

---

Companion to PR on uSockets: https://github.com/uNetworking/uSockets/pull/100

Thanks a lot for your work.

Signed-off-by: Aisha Tammy <gentoo@aisha.cc>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,9 @@
+DESTDIR ?=
+
+prefix ?= "/usr/local"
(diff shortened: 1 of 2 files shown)


alexhultman:
I've added a stripped down install target that simply copies the headers based on prefix, that should be enough for most users.

PR #1059, opened 2020-05-25
spelling typo

hst-m:


hst-m:
should just do a quick commit to fix these 2, doesnt need a PR

PR #1117, opened 2020-10-27
Feature cmake support

ithinx-io:


alexhultman:
I think your PR is nicely done overall, but I have no intention to add yet another "build system" when I know for a fact most people can do perfectly fine without such a thing.

I could write a 50 page book on why I don't want CMake, but that book would probably be "hate speech" so I'll just say this:

Striving for unity and "one build system to rule them all" is never in a billion years going to be a reality, so you'll just have to accept that different projects build in different ways. But this project is a header-only project so you don't even need any build system whatsoever. You only need to know how to link a static library to a C++ project.

DasRoteSkelett:
Sorry, @alexhultman , you are not serious on this, are you? We are not talking about "yet another" build system, we try to have at least one that is not a really mediocre hand crafted makefile, that will pose many problems when cross compiling, or using different setups. Using cmake (or meson, autotools, ...) is totally justified. If this project at least had one, this PR would have been totally unnecessary.

ithinx-io:
Hi @alexhultman,
Thanks for your feedback. I'm personally far away from voting CMake up or down, but as one of a widely used tool aiming to give (at least a little bit) support as a meta build approach capable of _generating_ Makefiles for you, it shouldn't be ignored that way.
Of course, there might be a lot of downsides. But in the end, the pros matter and saved my life with:

- Simplified setup
- Cross compiling environments
- Yocto integration

Seen from the perspective of components/modules where each one may have different options, licenses, etc.
it might be helpful keeping a complex project structure clean and sorted starting on git/repo level.
Mixing-up source repos (header copy, implicit includes, ...) looks like an easy way to start with, but a sustainable maintaining might become complicated one day.

Btw: Generating a header-only package using CMake happens the same way as one with sources needed to be compiled.

Long story short:
**CMake** support and recipes for **Yocto** are now available in the fork:
https://github.com/ithinx-io/uWebSockets

alexhultman:
I cross compile to 5 platforms daily without issues: https://github.com/uNetworking/uWebSockets.js/tree/binaries

Because this project relies rather on ISO standardized things like C and C++ to handle platform differences, there is nothing at all needed to build the project on any platform.

You include the files. That's it. No build system is required, no matter if you're building for Windows PC or Raspberry Pi - it's the same exact standard source code. You literally build uSockets.a, then you include the headers. That's it. It takes 5 minutes at most.

DasRoteSkelett:
When I try to build your project with yocto, uSockets.a is missing. So you need additional effort to add this to the yocto package. Again, if you prefer shared library over a static one, you have to touch the Makefiles. Every of you Options needs to be documented (which you have done) or you have to read the Makefile to see that there are any. If you do the standard `make DESTDIR=/my/prefix/path install` , you end up with `/my/prefix/path/usr/local/include`, which is not expected (the usr/local part of it).  
Personally, I will refrain from using your library. 

Regards

alexhultman:
> Personally, I will refrain from using your library.

> Regards

I understand that you believe you just did a "rad mic drop" here. The thing is - nobody cares. The world does not revolve around you and your opinions alone - there are about a thousand more people that have opinions here and their opinions are not about "does it or doesn't it work out of the box with Yocto". This is an important personality trait - to understand ones own limited importance. I could write a book about this after having dealt with random people on the internet for 5 years now. People need to understand the world is bigger than their own inner thoughts and their needs.

alexhultman:
For historic accuracy - the project did in fact have CMake back in 2016 but was removed because it didn't work out:

* a) People could not agree on anything so changes to the CMake files were by far the most popular PR topic (sometimes people sent 2 PRs for CMake per day and many PRs conflicted with others). I have no interest in resolving conflicts among fanatical "build systems"-people all day long.
* b) CMake is the lowest common denominator among all compilers and has limitations on flag accuracy.
* c) CMake introduces unnecessary dependencies and many have outdated versions.
* d) Most people who care about CMake are Windows users and Windows is a second-citizen target

Nowadays - people send me actual PRs for the actual source instead of meta-PRs for meta-subjects.

alexhultman:
> If you do the standard make DESTDIR=/my/prefix/path install , you end up with /my/prefix/path/usr/local/include, which is not expected (the usr/local part of it).

So set the prefix="" as per [standard GNU make documentation](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html). How did you navigate your legs into your pants? Honestly, if you cannot spend the 5 minutes it takes to figure things out (or read), your opinion becomes even less important.

Honestly, the more I try to listen to your opinion the more I reinforce my original take on "build systems". You try and solve problems that aren't there, by adding more shit that is unnecessary and will lead to even more problems in the long run.

All you really needed was `prefix="" DESTDIR=whatever make` and if you want to cross-compile, set the CC and CXX!

DasRoteSkelett:
Hi @alexhultman ,

seems I got you triggered. 
And lol, thanks for pointing out my role in the world. To me, I am still the most important person in the world ;-) 
Your points b-d are just wrong. You do not like cmake? Fine. cmake is far from perfect. But ruling out *any* build system tells me, you are living in the 1980s and you do not want to improve on tooling. 

Please leave a comment after this, so you can have the final word in this matter,

Regards

hst-m:
@DasRoteSkelett what an idiot gtfo you sound like small child

alexhultman:
> Please leave a comment after this, so you can have the final word in this matter,

This is a pointless attempt at manipulation - of course I will have the last word! This is my workplace I come here every day and I moderate everything here. In case you didn't realize - I run this project.

> To me, I am still the most important person in the world ;-)

You are the most important **to you**. To **me** you're a random guy on the internet demanding a feature that I know will bring friction to this project. (as evidently demonstrated in this very thread) And again, I run this project so I have to try and please as many as I can while still being true to the vision I have.

In other words, your opinion is insignificant. Insignificant but still valid - but insignificant.

PR #1119, opened 2020-11-09
fix access outside the array

ihsinme:
in the original version, access outside the array is possible. it seems to me that the code looks more secure.


alexhultman:
Nice catch! You're probably right - looks like a mistake. Thankfully this is not an issue in practice since the parser runs on a padded buffer. I will have a look and see if this makes sense to merge (it probably does but I'm not 100% sure right now).

PR #1124, opened 2020-11-12
1. Use the size_t to instead of the int to avoid the GCC warning.

pengweichu:
2. Initialize the struct Node and struct CloseFrame to avoid the GCC warning.
Since in my project, I treat the warning as error so can't get it compiled.


alexhultman-2fa:
Can you confirm latest master has these fixed? They should be fixed

PR #1116, opened 2020-10-26
Minor example updates

pavulon:
https://github.com/uNetworking/uWebSockets.js/issues/377

PR #1131, opened 2020-11-23
Fix TAD on old examples by adding .upgrade handler

### LINKED ISSUE #1130: Bad template deduction in some example files
`BroadcastingEchoServer`, `EchoServer`, and `EchoServerThreaded` fail to compile/deduce due to lacking the `.upgrade` handler on older versions of GCC.

EDIT: Seems to have been only partially fixed in PR #1116

Sunstradamus:
Fixes #1130 and follows up on #1116


alexhultman:
Okay this is not required for new compilers but let's merge it so that examples compile for more people.

PR #1137, opened 2020-12-06
Fix compilation error in HttpServer example

alastairpatrick:


alastairpatrick:
Tangentially related, for a future pull request, not this one, what do you think of something like this:

struct SocketContextOptions : us_socket_context_options_t {
    SocketContextOptions() : us_socket_context_options_t({}) {
    }
};

The idea is that the member variables would all be initialized to default values without the need to be kept in sync between two structs and there would be no need for a conversion operator since SocketContextOptions would inherit from us_socket_context_options_t.

alexhultman:
Yes SocketContextOptions was added very recently to fix a warning, it can be improved

PR #1154, opened 2020-12-24
Modified UpgradeAsync to use threads and loop->defer()

### LINKED ISSUE #1153: Calling websocket->close() and websocket->write() from another thread
Sorry for bothering again. I have another crash on Win64 but I'm not sure if this is a bug or wrong library usage.

Should it be safe to call `websocket->close()` from another thread (or `websocket->write()`) ? 

Occasionally the application is crashing after calling `websocket->close()` from the main application thread (when simulating manual client disconnection in our tests). 

It's because `s->context` is `0xddddddddddd` during `TopicTree::unsubscribeAll()`

But I'm not sure if this is a bug or it's caused by calling it from the other thread because this code https://github.com/uNetworking/uSockets/blob/master/src/socket.c#L65 doesn't seem to thread-safe.

Is there any list of methods that are safe to call from another thread? 

And is there any "writeable" callback in the uWebsockets thread where is it possible to run these not thread-safe methods? I read about `us_timer_set` which should be probably utilized too (https://github.com/uNetworking/uWebSockets/blob/master/examples/UpgradeAsync.cpp)

Thank you

Sunstradamus:
Resolves #1153; hopefully this also stops future questions about multithreading `ws->write()` as well.

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -5,7 +5,7 @@ on: [push]
 jobs:
   build_linux:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     
     steps:
     - name: Clone source
@@ -23,4 +23,5 @@ jobs:
       run: git clone --recursive https://github.com/uNetworking/uWebSockets.git
     - name: Build source
       run: make -C uWebSockets
-      
+    - name: List binaries
+      run: ls uWebSockets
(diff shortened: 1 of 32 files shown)


alexhultman:
Sure but "async" does not automatically mean "threaded". It would be a much better example if you instead of spawning a thread, made a timer using uSockets. That way you better demonstrate the single-threaded nature of the library.

alexhultman:
Wait what - I already had it that way. This PR is not an improvement IMO.

PR #1162, opened 2021-01-06
Fix potential memory leak in WebSocket

ShekMaha:
The `sendBuffer` created using `malloc` in `AsyncSocket.h` is not freed in all the paths returning from this function. This potential memory leak was identified by a clang-tidy tool check.


alexhultman:
The sendBuffer is only malloc:ed in case of requiresWrite being true. So in all other cases it shouldn't be free:d (those are the fast paths).

Issue #1132, opened 2020-11-27
Print warnings on not being able to load certificates


epsilon-0:
I have an SSL app, which ran but all SSL connections failed as the certificates were with incorrect permissions (system admin had messed up).

Knowing that the problem was with the loading of certificates would have been helpful to debug. It took us really long to figure out it was a file issue and not an OpenSSL version issue.  
Is there a way to have debug information with the library, so we can catch errors like this?

Thanks,
Aisha

alexhultman-2fa:
Didn't the listen function just fail?

epsilon-0:
No, it didn't fail. :open_mouth: 
Trying to read the files was causing an `ENOENT` error but the listen proceeded and the server was running.

joshxyzhimself:
Possibly related, just to add:

- I had an SSLApp silently fail (just returns empty response) when I accidentally used port `80`, instead of `443`.
- `ca_file_name` isn't mentioned on https://unetworking.github.io/uWebSockets.js/generated/interfaces/appoptions.html, but I found it at https://github.com/uNetworking/uWebSockets/blob/master/src/App.h#L38. It still worked though even without providing it (maybe not needed at all).

I'm on Ubuntu `20.04`, uWebSockets.js `v18.12.0`

joshxyzhimself:
Update: it looks like it only works on chrome desktop, but not on firefox desktop and chrome android.

I've tried the following but no luck:

```js
let app = null;
let port = null;

const endpoint_domain = 'mydomain.com';
const key_file_name = `/etc/letsencrypt/live/${endpoint_domain}/privkey.pem`;
const cert_file_name = `/etc/letsencrypt/live/${endpoint_domain}/cert.pem`;
const ca_file_name = `/etc/letsencrypt/live/${endpoint_domain}/chain.pem`;

if (fs.existsSync(key_file_name) === true) {
  assert(fs.existsSync(cert_file_name) === true);
  assert(fs.existsSync(ca_file_name) === true);

  app = uws.SSLApp({ key_file_name, cert_file_name, ca_file_name });
  port = 443;
} else {
  app = uws.App({});
  port = 8080;
}
```

alexhultman:
Maybe this should be tied up with logging

midlaneakali:
There also seems to be no warnings/errors if uWS is compiled without "WITH_OPENSSL=1" . It compiles and executes without using an ssl context even when the code tells it to.

joshxyzhimself:
[...]

PR #1163, opened 2021-01-06
add logging interface

jf99:
Currently, there is just a single logging line in uWebSockets which writes to `std::cerr`. I added two more such lines. More importantly, I introduced a function object `uWS::log` that can be overridden by users to accord with their logging library. For example, [spdlog](https://github.com/gabime/spdlog) users might want to write something like this:
```c++
uWS::log = [](const std::string& message, const int logLevel) -> void {
    switch(logLevel) {
        case 0:
            SPDLOG_ERROR(message);
            break;
        case 1:
            SPDLOG_WARN(message);
            break;
        case 2:
            SPDLOG_INFO(message);
            break;
        case 3:
            SPDLOG_DEBUG(message);
    }
};
```

Nothing changes for users who do not configure anything.

### DIFF
--- a/src/Log.h
+++ b/src/Log.h
@@ -34,7 +34,7 @@ class LogBuffer
         LogBuffer& operator<< (std::string_view sv) {
             const size_t bytesFree = buf.size() - cursor;
             if(sv.size() > bytesFree)
-                throw std::length_error("log message too long");
+                return *this;
             memcpy(&buf[cursor], sv.data(), sv.size());
             cursor += sv.size();
             return *this;
--- a/src/Log.h
+++ b/src/Log.h
@@ -13,9 +13,9 @@ namespace uWS {
  * \param [in] message The message to be logged.
  * \param [in] logLevel The severity of the message. 0 is error, 1 is warning, 2 is info. With each increment the severity decrements.
  */
-typedef std::function<void(const std::string& message, int logLevel)> LogFunction;
+typedef std::function<void(std::string_view message, int logLevel)> LogFunction;
 
-inline LogFunction log = [](const std::string& message, int logLevel) -> void {
+inline LogFunction log = [](std::string_view message, int logLevel) -> void {
     if(logLevel <= 1) {
         std::cerr << message << std::endl;
     }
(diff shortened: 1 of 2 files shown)


alexhultman:
Problem is that this PR introduces major performance regressions. Logging must not be added with a cost

jf99:
How about putting a macro around logging like this?
```c++
#if UWS_LOG_LEVEL <= 1
log("No handler for route " + std::string(httpRequest->getMethod()) + " " + std::string(httpRequest->getUrl()), 1);
#endif
```
The default value of `UWS_LOG_LEVEL` would be 0. That would let the user decide on the trade-off between performance and debuggability. spdlog does something [similar](https://github.com/gabime/spdlog/blob/v1.x/include/spdlog/spdlog.h#L243).

alexhultman:
Macro would work, but I don't want to put a bunch of text in the code - I would rather have it like so:

One header file contains all text -> you can either include it before App.h and then get logging enabled (or, let's say have WITH_LOG=1 build flag or something).

In the header you have macros like UWS_LOG_REQUEST(req) as functions, so that the calling (source code) remains clutter-free.

But it doesn't have to be C-style macros, you can use constexpr C++17 functions I mean you can have these "macros" as regular C++ code, but have an if constxpr (loglevel < 1) {

}

alexhultman:
Also, the logging has to be faster than just concatenating strings like that - use a pre-allocated array and memcpy or similar. One pre-allocated array per thread so mark it thread_local

jf99:
I have changed the code to fulfill your requirements:
- String concatenations are as efficient as possible
- One pre-allocated log buffer per thread
- No function call is made at all if the verbosity does not allow for messages of that loglevel
- Using the logger is a one-liner

<sup>It seemed overly complicated to me, but hey! My first real world problem solved with variadic templates.</sup>

I also made use of `if constexpr`. However a macro is still required, as parameters cannot be constexpr in C++17. The only way around this would be a template argument, I think, but I don't see how this is cleaner.

> One header file contains all text

Not sure if I understood you correctly, here. You want to move all log strings into a separate file like this?
```c++
const std::string noHandlerForRoute = "No handler for route ";
const std::string space = " ";
// ...
const std::string returningWithoutResponding = "Error: Returning from a request handler without responding or attaching an abort handler is forbidden!";
```
How would that be better?

src/Log.h
```diff
@@ -0,0 +1,72 @@
...
+class LogBuffer
+{
+    private:
+        std::array<char,1024> buf{0};
+        size_t cursor = 0;
+
+        LogBuffer& operator<< (std::string_view sv) {
+            const size_t bytesFree = buf.size() - cursor;
+            if(sv.size() > bytesFree)
+                throw std::length_error("log message too long");
```
jf99, line 37: You might also want to decide on this. Maybe it's more appropriate to silently return instead of throwing an exception and thereby potentially crashing the program.

jf99: After thinking more about it, throwing is a bad idea. Afaik you convert the C++ code to ASM.js for WebSocket.js, which I think doesn't work with exceptions.

jf99:
@alexhultman Could you please leave a comment on the changes that I've made? Thanks.

alexhultman:
#1132

jf99:
I added some more logging, including the case when `us_create_socket_context` fails. However, we cannot log any SSL specific stuff here, as this is handled by uSockets.

alexhultman:
Then maybe logging should be part of uSockets

jf99:
@alexhultman Do we still need logging in uSockets, now that you [added](https://github.com/uNetworking/uSockets/commit/e8990f6b0e48d14523417cfe6ca283f525343c51) an error handler for SSL? I might have some time in the next days.

PR #1165, opened 2021-01-14
Fixes compilation error due to a missing 'struct' keyword

milostosic:
us_socket_t type is normally preceded by 'struct' keyword, this is missing in one spot causing compile error.


milostosic:
I will create a new pull request with compiler warning fix as well.

PR #1166, opened 2021-01-14
Fixes a compilation error and warning under MSVC/VS2019

milostosic:
This minor change fixes a compilation error (missing 'struct' keyword) and warning (warning C4138: '*/' found outside of comment) while compiling with MSVC/VS2019.

### DIFF
--- a/src/AsyncSocket.h
+++ b/src/AsyncSocket.h
@@ -65,7 +65,7 @@ protected:
     }
 
     /* Immediately close socket */
-    us_socket_t *close() {
+    struct us_socket_t *close() {
         return us_socket_close(SSL, (us_socket_t *) this, 0, nullptr);
     }
 
--- a/src/WebSocketContextData.h
+++ b/src/WebSocketContextData.h
@@ -147,12 +147,12 @@ public:
         return 0;
     }) {
         /* We empty for both pre and post just to make sure */
-        Loop::get()->addPostHandler(this, [this](Loop */*loop*/) {
+        Loop::get()->addPostHandler(this, [this](Loop * /*loop*/) {
             /* Commit pub/sub batches every loop iteration */
             topicTree.drain();
         });
 
-        Loop::get()->addPreHandler(this, [this](Loop */*loop*/) {
+        Loop::get()->addPreHandler(this, [this](Loop * /*loop*/) {
             /* Commit pub/sub batches every loop iteration */
             topicTree.drain();
         });


alexhultman:
This doesn't add up. The missing struct is not a real issue - it is perfectly fine in C++ to not specify struct or class before - we do this at multiple places in the code.

Same with */ there are 91 places in the code where we do this (search for */*). This is a parser bug in Microsoft's precompiler.

Further on we compile on MSVC with every commit, producing binaries so I'm not sure about this.

milostosic:
I just tried with VS2017 as well and I'm seeing consistent errors/warnings. Maybe we have different warning level set? I am quite aggressive in my project configurations so maybe that's why I see this and you don't?
Here's the output from VS2017 build I attempted just now:

1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,17): error C2143:  syntax error: missing ';' before '*'
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,17): error C2143:     us_socket_t *close() {
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,17): error C2143:                 ^
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(260): message :  see reference to class template instantiation 'uWS::AsyncSocket<SSL>' being compiled
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(260): message : };
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,1): error C4430:  missing type specifier - int assumed. Note: C++ does not support default-int
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,1): error C4430:     us_socket_t *close() {
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,1): error C4430: ^
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,1): error C2334:  unexpected token(s) preceding '{'; skipping apparent function body
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,1): error C2334:     us_socket_t *close() {
1>e:\rtm\3rd\uwebsockets\src\asyncsocket.h(68,1): error C2334: ^
1>e:\rtm\3rd\uwebsockets\src\websocketcontextdata.h(150,55): warning C4138:  '*/' found outside of comment
1>e:\rtm\3rd\uwebsockets\src\websocketcontextdata.h(150,55): warning C4138:         Loop::get()->addPostHandler(this, [this](Loop */*loop*/) {
1>e:\rtm\3rd\uwebsockets\src\websocketcontextdata.h(150,55): warning C4138:                                                       ^
1>e:\rtm\3rd\uwebsockets\src\websocketcontextdata.h(155,54): warning C4138:  '*/' found outside of comment
1>e:\rtm\3rd\uwebsockets\src\websocketcontextdata.h(155,54): warning C4138:         Loop::get()->addPreHandler(this, [this](Loop */*loop*/) {
1>e:\rtm\3rd\uwebsockets\src\websocketcontextdata.h(155,54): warning C4138: 

These go away with the changes in my pull request.

alexhultman:
I will have a look at this. We recently fixed all pedantic macOS/Linux warnings so it could make sense to also fix all pedantic Windows warnings. But I still don't get why you only get warnings in AsyncSocket.h when we do the same things in other files too (91 occurrences of that comment-warning).

milostosic:
It's possible that once I actually start using the code I would see more warnings. At he moment it's just included. MSVC is weird that way with templates and tends to not spit out warnings if a template is not used. I can only guess this is some compilation time optimization for the compiler but I've seen it 'miss' warning before.

alexhultman:
I know with certainty these are not the only places where that comment-parsing-warning should come up. There should be close to 60 of those. I cannot merge this unless all occurrences are fixed

milostosic:
Well, maybe you should fix it all then.

alexhultman:
The only warnings I get are from microsofts own headers winbase.h. If I raise the warning level I get nonsense warnings. Not a priority as Windows is definitely a second-hand citizen for C++ dev

vekkuli:
If anyone else is running into these compilation errors, specifically including the

> error C2143: syntax error: missing ';' before '*'

I was able to reproduce this by including a more specific header than the **App.h** in a compilation unit. E.g., in a request handling class, I only included **HttpResponse.h**. I haven't looked into why this happens (perhaps someone else can explain?) but a fix is to simply include the App.h header in each compilation unit. Another option is not to produce separate compilation units and only use headers.

PR #1179, opened 2021-01-26
Fix checking if we actively drain backpressure

ronen37:
in WebSocketContext::us_socket_context_on_writable()


ronen37:
If I understand right, there is a mistake in checking if we actively drain backpressure.
Please verify my fix.

alexhultman:
Looks like you're right. Will check at this.

PR #1237, opened 2021-04-13
fix idleTimeout values 16,20,24,28 (8 second margin)

hst-m:


alexhultman-2fa:
Beat you to it 😄

hst-m:
One of these days I will learn enough c++ to print Hello World to console and maybe add a couple numbers together, not there yet though. Thankfully I can read it just enough to make out what is happening

alexhultman-2fa:
😉

PR #1240, opened 2021-04-14
Allow passing any functor to cork; no need for type erasure here

JWalker1995:


alexhultman-2fa:
What is the gain in having different callbacks for certain cases? Compiling with LTO produces the same result.

JWalker1995:
I didn't realize the compiler was that smart. Feel free to close.

alexhultman-2fa:
I pasted an example in compilerexplorer and it did actually not compile down to nothing - but with "function2" (another implementation) it did. So technically I'm wrong, but a solution would be to revert back to using function2 rather than put in specialized cases.

alexhultman-2fa:
I tested a more practical case, such as with cork - it does constant fold to zero cost in that case :)

Issue #1245, opened 2021-04-23
Shorter socket-timeouts after WebSocket::end()


not-implemented:
Currently after calling `WebSocket::end()`, the normal idleTimeout (default 120s) is active. For broken/hanging connections (reproducable with iptables DROP of WebSocket port after connecting), the socket is still active until this timeout.

Normally this is not a big issue ... but it is, when doing a graceful shutdown of the whole application. Such broken clients cause the event loop to hang until all the timeouts are reached.

One solution will be, to call `WebSocket::close()` instead of `end()` on shutdown, it works ... but it is not "graceful" anymore, of course ;-)

Another solution will be, to expose `WebSocket::timeout()` as public, but maybe you don't want this, because it will interfere with the internal timeout handling and will be very misunderstanding.

The third solution will be, to set a shorter timeout in `WebSocket::end()` ... for example to `webSocketContextData->idleTimeoutComponents.second` which you used for the auto-ping-timeouts. (`WebSocketContextData::calculateIdleTimeoutCompnents` needs to be adjusted, so `second` is also set if `sendPingsAutomatically` is off.)

Of course, I would prefer to have it configurable separately ... e.g. `pingTimeout` ... sometimes, too much "magic" is intransparent ;-)

What do you think? If you like, I can try a pull-request.

alexhultman:
Third option is nice imo

not-implemented:
Ok cool, I'll try to implement in the next days.

By the way - I don't know if you receive notifications on closed issues - did you see my comments in https://github.com/uNetworking/uSockets/pull/97 ? Unfortunately I broke the timeout implementation there.

alexhultman-2fa:
I saw but no release has that yet

PR #1246, opened 2021-04-26
Shorter socket-timeouts after WebSocket::end() (#1245)

not-implemented:
Just using the ping-timeout as end-timeout for now.

Setting the timeout after sending the close-message should also overwrite the timeout set by `send` if `resetIdleTimeoutOnSend` is active.

Also check `webSocketData->isShuttingDown` in `on_timeout()` ... this seems to be a small bug in any case (a ping may have been sent on a socket on timeout after being `shutdown()`).

### DIFF
--- a/src/WebSocket.h
+++ b/src/WebSocket.h
@@ -167,10 +167,14 @@ public:
             }
         }
 
-        /* Emit close event */
         WebSocketContextData<SSL, USERDATA> *webSocketContextData = (WebSocketContextData<SSL, USERDATA> *) us_socket_context_ext(SSL,
             (us_socket_context_t *) us_socket_context(SSL, (us_socket_t *) this)
         );
+
+        /* Set shorter timeout (use ping-timeout) to avoid long hanging sockets after end() on broken connections */
+        Super::timeout(webSocketContextData->idleTimeoutComponents.second);
+
+        /* Emit close event */
         if (webSocketContextData->closeHandler) {
             webSocketContextData->closeHandler(this, code, message);
         }
--- a/src/WebSocketContext.h
+++ b/src/WebSocketContext.h
@@ -364,7 +364,7 @@ private:
             auto *webSocketData = (WebSocketData *)(us_socket_ext(SSL, s));
             auto *webSocketContextData = (WebSocketContextData<SSL, USERDATA> *) us_socket_context_ext(SSL, us_socket_context(SSL, (us_socket_t *) s));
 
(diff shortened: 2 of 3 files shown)


alexhultman-2fa:
I don't even remember how the shutdown timeout works. Maybe it is better to have it configurable after all? Can you describe the bug in more detail?

What was the prior behavior, what is the new PR:ed behavior?

not-implemented:
Yes of course: When shutting down a socket, it, of course, wait's for the FIN-ACK of the peer before it get closed ... but with the current code, there is no different timeout set an the `idleTimeout` is still used then. As described in #1245 I have the following test-setup:

- start a WebSocket server with default 120s idleTimeout
- connect one client
- `iptables -A INPUT -p tcp --dport 8000 -j DROP` ... on the websocket port
- gracefully shutting down the WebSocket server (calling end() for each open connection - as there is no server.close() like in uWebSockets 0.14, right?)
- event loop hangs for 120s now until the application can exit, which is very long

What I expect is any possibility to set a shorter timeout here to optimize application restarts/shutdowns ... and yes, of course it will be better to have it configurable, but with the current interface, new options are always a breaking change, and I didn't mean to be too hasty ... but of course, something like a `pingTimeout` would be nice, or do you even mean 2 options like `pingTimeout` and `shutdownTimeout`? 

Regarding the bug: I'm not 100% sure, but I guess the following:

- assume `sendPingsAutomatically = true`
- call `end()` on a socket - it [send's the closePayload and calls `shutdown()` on the socket](https://github.com/uNetworking/uWebSockets/blob/master/src/WebSocket.h#L154-L168)
- assume a broken connection without getting a FIN-ACK
- The `us_socket_context_on_timeout()` callback will fire after `idleTimeout`
- The [`hasTimedOut`](https://github.com/uNetworking/uWebSockets/blob/master/src/WebSocketContext.h#L367) flag is never set outside of this function, so it must still be false and enter the "if"
- The [ping is sent](https://github.com/uNetworking/uWebSockets/blob/master/src/WebSocketContext.h#L371) on an already shutdown'ed socket ... bad

I found this primarily because the timeout is also reset here, which would be bad for this pull-request ... but it seems to be a bug in any case.

alexhultman-2fa:
You have to fix the broken timeout in uSockets before submitting PRs elsewhere

alexhultman-2fa:
Wrong button

not-implemented:
Cool, thanks!

PR #1249, opened 2021-05-03
Fix typos in READMORE

SuspiciousLookingOwl:


misc/READMORE.md
```diff
@@ -196,7 +196,7 @@ App.listen(port, [](auto *listenSocket) {
 })
 ```
 
-Cancelling listenning is done with the uSockets function call `us_listen_socket_close`.
+Cancelling listening is done with the uSockets function call `us_listen_socket_close`.
```
SuspiciousLookingOwl, line 199: Should `listening` be changed to `listener`?

hst-m: Listening is fine but Cancelling should be Canceling in the American English form, which we all know is the more correct form, less useless letters. Also you missed "toghether" on line 209 should be together

misc/READMORE.md
```diff
@@ -170,7 +170,7 @@ Similarly to for Http, methods such as ws.send(...) can cause backpressure. Make
...
 
 #### Backpressure
 Sending on a WebSocket can build backpressure. WebSocket::send returns an enum of BACKPRESSURE, SUCCESS or DROPPED. When send returns BACKPRESSURE it means you should stop sending data until the drain event fires and WebSocket::getBufferedAmount() returns a reasonable amount of bytes. But in case you specified a maxBackpressure when creating the WebSocketContext, this limit will automatically be enforced. That means an attempt at sending a message which would result in too much backpressure will be cancelled and send will return DROPPED. This means the message was dropped and will not be put in the queue. maxBackpressure is an essential setting when using pub/sub as a slow receiver otherwise could build up a lot of backpressure. By setting maxBackpressure the library will automatically manage an enforce a maximum allowed backpressure per socket for you.
```
hst-m, line 176: If you are changing these "cancelled" here should be "canceled", then you are American English approved 
```suggestion
Sending on a WebSocket can build backpressure. WebSocket::send returns an enum of BACKPRESSURE, SUCCESS or DROPPED. When send returns BACKPRESSURE it means you should stop sending data until the drain event fires and WebSocket::getBufferedAmount() returns a reasonable amount of bytes. But in case you specified a maxBackpressure when creating the WebSocketContext, this limit will automatically be enforced. That means an attempt at sending a message which would result in too much backpressure will be canceled and send will return DROPPED. This means the message was dropped and will not be put in the queue. maxBackpressure is an essential setting when using pub/sub as a slow receiver otherwise could build up a lot of backpressure. By setting maxBackpressure the library will automatically manage an enforce a maximum allowed backpressure per socket for you.
```

alexhultman-2fa:
Can you also fix the grammatical issues - are/is

SuspiciousLookingOwl:
> Can you also fix the grammatical issues - are/is

@hst-m can you help me with that? Not a native English speaker so my grammar probably isn't the best.

misc/READMORE.md
```diff
@@ -235,13 +235,13 @@ In the 1970s, programming was an elite's task. Today programming is done by uned
...
 
-Compression should be seen as a last resort, a temporary duckt-tape solution for when you cannot sit down and consider something better. Designing clever, binary and minimally repetitive protocols are going to save enormous amounts of CPU-time oterhwise lost to compression.
+Compression should be seen as a last resort, a temporary duck-tape solution for when you cannot sit down and consider something better. Designing clever, binary and minimally repetitive protocols are going to save enormous amounts of CPU-time otherwise lost to compression.
 
 In essence, compression is really just dynamically scanning for repetitions and gradually building up a dynamic palette of commonly repetitive chunks. That takes a lot of CPU-time and in incredibly inefficient. Overall, you're looking at only 20-30% remaining I/O performance if you use compression. Instead of letting some generic dynamic algorithm scan your inefficient data representation, you could have taken the time to design something that didn't suck in the first place.
```
hst-m, line 240:
```suggestion
In essence, compression is really just dynamically scanning for repetitions and gradually building up a dynamic palette of commonly repetitive chunks. That takes a lot of CPU-time and is incredibly inefficient. Overall, you're looking at only 20-30% remaining I/O performance if you use compression. Instead of letting some generic dynamic algorithm scan your inefficient data representation, you could have taken the time to design something that didn't suck in the first place.
```

alexhultman-2fa:
> Designing clever, binary and minimally repetitive protocols are going to save enormous amounts of CPU-time oterhwise lost to compression.

This sentence is grammatically wrong

alexhultman-2fa:
https://en.wikipedia.org/wiki/Duct_tape

PR #1253, opened 2021-05-06
Fix 4Gb limit streamed HTTP on Raspberry Pi 32bit

dmitigr:
Commits f2198ccfac140798ee066c5a3c5235b396473205 and 4fcfe49bcce574c0aca7ab6409f81a725bf8f88c
are fixed 2Gb limit, but introduced 4Gb limit on Rasperry Pi 32 bit.
This is because the max value of type `size_t` is `4294967295` on this
system.
Furthermore, the Standard Library use `std::uintmax_t` as a type
of value returned by `std::filesystem::file_size`.

### DIFF
--- a/src/HttpResponse.h
+++ b/src/HttpResponse.h
@@ -93,7 +93,7 @@ private:
 
     /* Returns true on success, indicating that it might be feasible to write more data.
      * Will start timeout if stream reaches totalSize or write failure. */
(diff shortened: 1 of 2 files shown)


alexhultman-2fa:
Good motivation - I checked with clang and size_t and intmax_t are both 8 byte on x64 so this change should only affect 32-bit?

dmitigr:
> so this change should only affect 32-bit?

Yes.

PR #1277, opened 2021-06-27
update forSubscriber

hst-m:
file: [TopicTree.h](https://github.com/uNetworking/uWebSockets/blob/86415757e7df3417754882dc6079d81ecb5e3408/src/TopicTree.h#L85-L148)
files changed: [Split view - Hide whitespace](https://github.com/uNetworking/uWebSockets/pull/1277/files?diff=split&w=1)

- Break outer For loop early if inner For loop finishes
- Add fast path for non-sending subscribers

### DIFF
--- a/src/TopicTree.h
+++ b/src/TopicTree.h
@@ -91,21 +91,20 @@ struct Intersection {
 
         std::pair<size_t, size_t> toEmit = {};
         std::pair<size_t, size_t> emitted = {};
-        unsigned int startAt = 0;
+        unsigned int examinedHoles = 0;
 
         /* Iterate each message looking for any to skip */
         for (auto &message : holes) {
 
             /* If this message was sent by this subscriber skip it */
             bool skipMessage = false;
-            for (unsigned int i = startAt; i < senderForMessages.size(); i++) {
-                if (senderForMessages[i] > message.messageId) {
-                    startAt = i;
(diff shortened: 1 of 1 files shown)


alexhultman:
Calm down a bit. This is negligible and not important.

Latest PR needs fixing - you swapped:

for senderForMessages:
for holes

to

for holes:
for senderForMessages

This is a performance regression. Now you seek for every global hole, for every local sender, This means ALL sockets take a performance hit FOR ALL HOLES IN THE WHOLE TREE.

Compare this with ONLY SENDERS take a hit for all holes in the whole tree.

hst-m:
Holes = messages in this intersection

"For" loops for Holes and senderForMessages needed to be switched to fix bug, but they both still iterate over the same total number of items, so not sure what you are referring to, I will look some more

alexhultman:
Hmm, my bad.. I think. But I am thinking about a completely rewritten TopicTree without the nonStrict bug/fix.

hst-m:
Could probably rename some variables to make it more clear, you may have made updates that changed their meaning, like "holes" should prob be "messages" because they are all the messages in that intersection, but is not priority

hst-m:
That outer "For" loop could be made more efficient by breaking out early when inner loop was done so I added that

PR #1289, opened 2021-07-20
Fix for case - when server subscribed to to topic and connection is closed segmentation fault will happen

vlovch01:
Steps to reproduce:
I have a multi-threaded server where each thread is having it's own websocket ( message loop, ...).
When a connection is happen I am subscribing to a topic.
After connection is closed from client side server is trying to clean the subscribers but a segmentation fault is happening.

I found that the reason for this is that no proper clean up is done after memory is de-allocated.

In file _WebSocketContext.h_ when socket is closed there is a call for 
            **webSocketData->~WebSocketData();**
            
Destructor will delete the allocated memory. 
Then in **Websocket.h**  will be called **end** function which will call  _unsubscribeAll_ . 
In side unsubscribeAll check for if subscriber is nullptr will pass but next invocation for 
**subscriber->subscriptions** will create seg fault because memory was already deleted in destructor. 

Reason for this  seg fault of explicit call of destructor **webSocketData** in init method **WebSocketContext.h**
This call looks like should happen later.

Moved explicit call of destructor to end function in Websocket.h.

### DIFF
--- a/src/TopicTree.h
+++ b/src/TopicTree.h
@@ -214,8 +214,11 @@ private:
             }
 
             /* Free various memory for the node */
+            topic->fullName.clear();
             delete [] topic->name;
+            topic->name = nullptr;
             delete topic;
+            topic = nullptr;
 
             if (parent == root) {
               break;
--- a/src/TopicTree.h
+++ b/src/TopicTree.h
@@ -183,7 +183,7 @@ private:
     Subscriber *min = (Subscriber *) UINTPTR_MAX;
 
     /* Cull or trim unused Topic nodes from leaf to root */
-    void trimTree(Topic *topic) {
+    void trimTree(Topic *&topic) {
         while (!topic->subs.size() && !topic->children.size() && !topic->terminatingWildcardChild && !topic->wildcardChild) {
             Topic *parent = topic->parent;
 
@@ -456,9 +456,9 @@ public:
     }
 
     /* Can be called with nullptr, ignore it then */
-    void unsubscribeAll(Subscriber *subscriber, bool mayFlush = true) {
+    void unsubscribeAll(Subscriber *&subscriber, bool mayFlush = true) {
         if (subscriber) {
-            for (Topic *topic : subscriber->subscriptions) {
+            for (auto& topic : subscriber->subscriptions) {
 
                 /* We do not want to flush when closing a socket, it makes no sense to do so */
 
--- a/src/WebSocket.h
+++ b/src/WebSocket.h
@@ -179,6 +179,8 @@ public:
         webSocketContextData->topicTree.unsubscribeAll(webSocketData->subscriber, false);
         delete webSocketData->subscriber;
         webSocketData->subscriber = nullptr;
+        webSocketData->~WebSocketData();
+        webSocketData = nullptr;
     }
 
     /* Corks the response if possible. Leaves already corked socket be. */
--- a/src/WebSocketContext.h
+++ b/src/WebSocketContext.h
@@ -254,9 +254,6 @@ private:
                 webSocketData->subscriber = nullptr;
             }
 
-            /* Destruct in-placed data struct */
-            webSocketData->~WebSocketData();
-
             return s;
         });
 
--- a/src/WebSocketData.h
+++ b/src/WebSocketData.h
@@ -62,10 +62,12 @@ public:
     ~WebSocketData() {
         if (deflationStream) {
             delete deflationStream;
+            deflationStream = nullptr;
         }
 
         if (subscriber) {
             delete subscriber;
+            subscriber = nullptr;
         }
     }
 };


alexhultman:
This is not a correct analysis and this fix is not correct. You most likely have done something invalid wow using multiple threads.

Please provide a snippet that reproduces the bug you claim happens.

alexhultman-2fa:
Closing PR, feel free to report bugs if you can pinpoint anything

alexhultman-2fa:
> When a connection is happen I am subscribing to a topic.

> After connection is closed from client side server is trying to clean the subscribers but a segmentation fault is happening.

This is what BroadcastingEchoServer example does and I just tested connecting to it, disconnecting from client side and it had no issues, running under AddressSanitizer.

This bug report is of something that is very fundamental, a smoke test. So for it to broken like you describe, I just don't buy it without extensive proof.

vlovch01:
> 
> 
> > When a connection is happen I am subscribing to a topic.
> 
> > After connection is closed from client side server is trying to clean the subscribers but a segmentation fault is happening.
> 
> This is what BroadcastingEchoServer example does and I just tested connecting to it, disconnecting from client side and it had no issues, running under AddressSanitizer.
> 
> This bug report is of something that is very fundamental, a smoke test. So for it to broken like you describe, I just don't buy it without extensive proof.

looks like I got what is the problem.
I have in my code in lambda close the call **ws->close();** 
because of this code the **init** method is called first and the the **end** 
which causes seg fault. 
It is reproducing stable on simpler example just add in close lambda call ws->close();
and client needs to sent the close connection request.

By removing the ws->close(); seg fault is not reproducing.
But any way you need to clean up the code after deleting the memory assign nullptr to pointers.


Here is the seg fault with valgrind:
==22826== Invalid read of size 8
==22826==    at 0x11F31B: std::__cxx11::list<uWS::Topic*, std::allocator<uWS::Topic*> >::begin() (stl_list.h:946)
==22826==    by 0x11B080: **uWS::TopicTree::unsubscribeAll(uWS::Subscriber*&, bool) (TopicTree.h:449)**
==22826==    by 0x13EF03: uWS::WebSocket<false, true, PerSocketData>::end(int, std::basic_string_view<char, std::char_traits<char> >) (WebSocket.h:179)
==22826==    by 0x13BC3A: uWS::WebSocketContext<false, true, PerSocketData>::handleFragment(char*, unsigned long, unsigned int, int, bool, uWS::WebSocketState<true>*, void*) (WebSocketContext.h:168)
==22826==    by 0x136E83: bool uWS::WebSocketProtocol<true, uWS::WebSocketContext<false, true, PerSocketData> >::consumeMessage<6u, unsigned char>(unsigned char, char*&, unsigned int&, uWS::WebSocketState<true>*, void*) (WebSocketProtocol.h:328)
==22826==    by 0x13306B: uWS::WebSocketProtocol<true, uWS::WebSocketContext<false, true, PerSocketData> >::consume(char*, unsigned int, uWS::WebSocketState<true>*, void*) (WebSocketProtocol.h:424)
==22826==    by 0x12CE8F: _ZZN3uWS16WebSocketContextILb0ELb1E13PerSocketDataE4initEvENKUlPT_PciE0_clI11us_socket_tEEDaS4_S5_i (WebSocketContext.h:286)
==22826==    by 0x137BF6: _ZZN3uWS16WebSocketContextILb0ELb1E13PerSocketDataE4initEvENUlPT_PciE0_4_FUNI11us_socket_tEEDTcldtdeLKPKS6_0EonclIS3_EscOS4_fp_scOS5_fp0_scOifp1_EES4_S5_i (WebSocketContext.h:264)
==22826==    by 0x1482A6: us_internal_dispatch_ready_poll (loop.c:275)
==22826==    by 0x147196: us_loop_run (epoll_kqueue.c:147)
==22826==    by 0x1186C9: uWS::Loop::run() (Loop.h:159)
==22826==    by 0x1186E6: uWS::run() (Loop.h:176)
==22826==  Address 0x519dc10 is 0 bytes inside a block of size 32 free'd
==22826==    at 0x4845B0F: operator delete(void*, unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==22826==    by 0x11BF22: uWS::WebSocketData::~WebSocketData() (WebSocketData.h:69)
==22826==    by 0x12CDA9: _ZZN3uWS16WebSocketContextILb0ELb1E13PerSocketDataE4initEvENKUlPT_iPvE_clI11us_socket_tEEDaS4_iS5_ (WebSocketContext.h:258)
==22826==    by 0x136B49: _ZZN3uWS16WebSocketContextILb0ELb1E13PerSocketDataE4initEvENUlPT_iPvE_4_FUNI11us_socket_tEEDTcldtdeLKPKS6_0EonclIS3_EscOS4_fp_scOifp0_scOS5_fp1_EES4_iS5_ (WebSocketContext.h:240)
==22826==    by 0x148822: us_socket_close (socket.c:109)
==22826==    by 0x121322: uWS::AsyncSocket<false>::close() (AsyncSocket.h:69)

alexhultman-2fa:
I can reproduce the problem by calling ws->close() in the close handler. It requires having subscribed to a topic. But this is not really a bug - you are breaking the rule that a socket is valid only from open event to close event -> by USING the socket in close handler you are using a resource that is no longer valid. A check to catch this misuse could be added, and I know that is the design used elsewhere but in this case, it ultimately boils down to misuse.

alexhultman-2fa:
A fix could be to check if isShuttingDown is true, in WebSocket::close handler. This bool is true when shut down from client.

vlovch01:
Yes I have a subscription to topic for the socket as soon as connection is opened.

But yeah having ws->close() in close lambda needs to be somehow handled because it's just crash the server. 
And this error is very easy to do.

vlovch01:
> 
> 
> A fix could be to check if isShuttingDown is true, in WebSocket::close function. This bool is true when shut down from client.

I will check that.

PR #1335, opened 2021-09-20
Update Node.h

LazyHippogriff:
In case options has uS::ONLY_IPV4 set, the present handling is not enclosed within 'else' block.


alexhultman:
0.14 branch is deprecated 4 years ago and sees no patching or development here.

PR #1399, opened 2022-01-13
Add CompressionIOS15.cpp example

ewak:
Make build work on centos7 with devtoolset-9 and openssl11
Fix passing of compiler and linker flags
Add testcerts makefile target


alexhultman:
Kind of the same as I said in the issue; you have a lot of energy and that is good but the goal of this project is not to adapt to Apple bugs but rather follow spec. So adding per-Apple bug examples is not a goal

ewak:
Sure no problem.  

A PR was just an easy way to share a complicated reproduction of the issue.

PR #1401, opened 2022-01-16
Fix possibly race condition in Loop.h

summer-espresso:
Hello, thank you for your library.
I experienced crashes due to segfault and double-free corruption.  
After this patch, I didn't have any more crashes (although I'm still not totally sure I use this patch)
That's why I submit you this pull request in case you think it's relevant and doesn't burden the performance too much.


alexhultman:
This PR introduced dead locks and does not solve anything. There is no issue; there are two queues and they alternate as the loop iterates. You have issues elsewhere.

summer-espresso:
@alexhultman Thanks for your comment. I will investigate my code further.

Issue #1347, opened 2021-09-28
iOS 15 and compression is broken


alexhultman:
**All versions of uWS** are _totally_ incompatible with iOS 15 when using compression and will not keep a connection. This is because of an Apple bug where client_no_context_takeover is not properly implemented in newest Safari for iOS 15.

Because of this, if you use compression and expect iOS 15 connections, you **will have to** enable uWS::DEDICATED_DECOMPRESSOR as well.

See EchoServer.cpp example. uWS::DEDICATED_DECOMPRESSOR is available in v20.3.0 and forward. This adds about 44kb of memory usage per connection.

alexhultman:
This bug is also in Safari 15 on macOS 12. I have reported this to Apple but it's impossible to know if they will ever read it or forward to nearest trash can.

alexhultman:
Safari does NOT work with the new DEDICATED_DECOMPRESSOR_xxx but MUST ONLY be used with the full DEDICATED_DECOMPRESSOR.

The reason is because Safari literally does not support client_max_window_bits, so uWS falls back (according to standard) to setting the client_no_context_takeover which Safari does not support (which it MUST support, according to standard). Then it fails with protocol error.

This sucks because it means the only viable setting is DEDICATED_DECOMPRESSOR, unless some kind of per-Safari hack is added (like we already had to do with Safari's x-webkit-deflate-frame before).

not-implemented:
Thanks for documenting this ... we ran into the same issue :-(

It really sucks ... as using `DEDICATED_DECOMPRESSOR` is not an option for us because of the massive RAM usage for ten thousands of connections ...

Would be nice if you can keep us up2date if there are news on the apple issue.

alexhultman:
Well it's Apple so it cannot possibly be their fault.

not-implemented:
Still no news from Apple here, right? That's really annoying ;-)

alexhultman:
Apple feedback assistant:

![](https://thumbs.gfycat.com/UnhappyFatherlyGoa-max-1mb.gif)

ewak:
Here is what I think is the relevant issue in the webKit bugzilla

REGRESSION (iOS 15): Websocket connection instance in javascript client getting closed
https://bugs.webkit.org/show_bug.cgi?id=228296#c37

And a comment that gives ?hope? of an apple response.
> Alex Christensen 2021-12-16 19:59:07 PST
[...]

PR #1414, opened 2022-02-02
Updated Safari hack to target 15.0-15.3 (#1347)

not-implemented:
To be merged when we are 100% sure, that Safari 15.4 is fixed.


alexhultman:
Nice, I merged it we can change the version number if it doesn't work. Master is dev anyways

PR #1436, opened 2022-03-28
fix compile error

BugLesser:
error message: "Mismatching uSockets/uWebSockets ABI".
cause: struct 'us_socket_context_options_t' has no 'ssl_ciphers' member variable.

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -76,7 +76,6 @@ namespace uWS {
         const char *passphrase = nullptr;
         const char *dh_params_file_name = nullptr;
         const char *ca_file_name = nullptr;
-        const char *ssl_ciphers = nullptr;
         int ssl_prefer_low_memory_usage = 0;
 
         /* Conversion operator used internally */


ariestav:
I'm installing uWebSockets via vcpkg like so:

`vcpkg install uwebsockets:x64-windows`

Then to build with Visual Studio 2022, I setup the "Additional Includes Directories" to have the `/include` directory that `vcpkg` adds to the `installs` directory.  I get the ABI mismatch error, but if I put back in the line of code referenced in this commit, then I don't get the error and the build succeeds.

Is there something to do to ensure that the vcpkg installer is getting the latest build?

alexhultman:
I don't think vcpkg uses the latest at all, they manually update it with certain commits and releases. No idea what versions work, or don't work. Try and get everything from vcpkg and nothing from GitHub and you should probably be fine.

ariestav:
That seems to do the trick, thanks.  Who manages the repos that vcpkg retrieves the code from?  Is there a way to find out the version that vcpkg is using during install?

alexhultman:
Microsoft does that, no idea what versions they use. Should be printable in some way.

PR #1449, opened 2022-04-26
re-add custom ssl_ciphers support

shaharmor:
Follow up of https://github.com/uNetworking/uSockets/pull/179
Fixes https://github.com/uNetworking/uWebSockets/issues/1437

PR #1450, opened 2022-04-26
update uSockets

mumbleskates:
This change, and the corresponding submodule update, should make uWebSockets compatible with newer versions of uSockets that now include the `ssl_ciphers` SSL socket context option.

see: https://github.com/uNetworking/uSockets/issues/173, https://github.com/uNetworking/uSockets/pull/179


mumbleskates:
ope: https://github.com/uNetworking/uWebSockets/pull/1449

PR #1486, opened 2022-09-26
add unix domain socket support for websocket

Young-Flash:


uNetworkingAB:
You get a gold star ⭐

ddevienne:
Having no experience with those, I'm just wondering if there's a performance advantage to using them?
Or some other quality that makes them preferable in some contexts? Like unit testing perhaps?

uNetworkingAB:
Pretty sure the only common use case is if you have a proxy locally on the same machine (which should be pretty rare?).

PR #1495, opened 2022-10-18
Merge bun changes

uNetworkingAB:


Jarred-Sumner:
oh i didn't see this

cool

uNetworkingAB:
Yeah I gradually take parts from here. But GitHub does not update its diff

benchmarks/load_test.c
```diff
@@ -1,7 +1,29 @@
 /* This is a simple yet efficient WebSocket server benchmark much like WRK */
 
 #define _BSD_SOURCE
+
+#ifdef __APPLE__
+#include <libkern/OSByteOrder.h>
+
+#define htobe16(x) OSSwapHostToBigInt16(x)
```
Jarred-Sumner, line 8: rationale for this was to make `load_test` work on macOS

macOS is missing these functions

examples/EchoServer.cpp
```diff
@@ -12,11 +12,17 @@ int main() {
 
     /* Keep in mind that uWS::SSLApp({options}) is the same as uWS::App() when compiled without SSL support.
      * You may swap to using uWS:App() if you don't need SSL */
-    uWS::SSLApp({
+    uWS::App({
```
Jarred-Sumner, line 15: this was so I could run the EchoServer without HTTPS, but there was no need to really commit this

src/App.h
```diff
@@ -153,6 +153,13 @@ struct TemplatedApp {
         httpContext->filter(std::move(filterHandler));
     }
 
+    /* Publishes a message to all websocket contexts - conceptually as if publishing to the one single
+     * TopicTree of this app (technically there are many TopicTrees, however the concept is that one
+     * app has one conceptual Topic tree) */
+    bool publish(std::string_view topic, std::string_view message, unsigned char opCode, bool compress = false) {
```
Jarred-Sumner, line 159: this is dumb and as discussed shouldn't include it

src/ChunkedEncoding.h
```diff
@@ -29,12 +29,12 @@
 
 namespace uWS {
 
-    uint32_t STATE_HAS_SIZE = 0x80000000;
-    uint32_t STATE_IS_CHUNKED = 0x40000000;
-    uint32_t STATE_SIZE_MASK = 0x3FFFFFFF;
+    constexpr uint32_t STATE_HAS_SIZE = 0x80000000;
```
Jarred-Sumner, line 32: this was so we could include uWS in two different object files

In one, it's the C bindings

In the other, it's to copy the `Headers` object from JS bindings -> uWS response object

There were duplicate symbol errors without `constexpr` / `inline` / `static`

src/Loop.h
```diff
@@ -79,6 +80,8 @@ struct Loop {
         /* We also need some timers (should live off the one 4 second timer rather) */
         LoopData *loopData = (LoopData *) us_loop_ext((struct us_loop_t *) loop);
         loopData->dateTimer = us_create_timer((struct us_loop_t *) loop, 1, sizeof(LoopData *));
+        loopData->updateDate();
```
Jarred-Sumner, line 83: This was to ensure we always have a date set

.gitmodules
```diff
@@ -1,9 +1,12 @@
 [submodule "uSockets"]
-	path = uSockets
-	url = https://github.com/uNetworking/uSockets.git
+path = uSockets
+url = https://github.com/Jarred-Sumner/uSockets
```
Jarred-Sumner, line 3: You probably don't want to commit this

src/Loop.h
uNetworkingAB: It's set in the constructor

src/ChunkedEncoding.h
uNetworkingAB: this whole part is already included (but github diff doesn't update)

src/App.h
uNetworkingAB: right :)

examples/EchoServer.cpp
uNetworkingAB: yep, examples won't change

benchmarks/load_test.c
uNetworkingAB: already included

.gitmodules
uNetworkingAB: you can checkout your variant until uSockets is merged

N0N1m3 (approved):

PR #1509, opened 2022-11-02
Use ULL (unsigned long long) instead of UL

rehans:
The HelloWorld example failed on Windows (with MSVC) whereas it succeeded to run on macOS (XCode). So debugging both platforms in parallel revealed the parsing error due to using '0UL' instead of '0ULL'. 

I checked both platforms with '0ULL' successfully now though I did not check Linux.

### DIFF
--- a/src/HttpParser.h
+++ b/src/HttpParser.h
@@ -214,15 +214,15 @@ private:
     }
     
     static inline uint64_t hasLess(uint64_t x, uint64_t n) {
-        return (((x)-~0UL/255*(n))&~(x)&~0UL/255*128);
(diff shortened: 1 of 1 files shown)


uNetworkingAB:
Very nice find. I was going insane at the Windows test failing, now it works :)

rehans:
Ah, nice that it works now! :)

uNetworkingAB:
Wow now the entire test works on all platforms, even the full "smoke test" (really a small integration test)

rehans:
LOL Great! :D

PR #1510, opened 2022-11-02
Move TopicTreeMessage and TopicTreeBigMessage to TopicTree.h

jwoos:
`TopicTreeMessage` and `TopicTreeBigMessage` definitions are required prior to including App.h since I previously (#1507) changed it to do a forward declaration. The forward declaration created issues as other parts ended up accessing the structs, which require the definitions.

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -23,18 +23,6 @@
 #include <string_view>
 
 namespace uWS {
-    /* Type queued up when publishing */
-    struct TopicTreeMessage {
-        std::string message;
-        /*OpCode*/ int opCode;
-        bool compress;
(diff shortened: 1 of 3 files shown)


uNetworkingAB:
It can't live in topic tree since that's the template class. Topic tree should not provide some definition of T

jwoos:
How do you feel about moving it out into a separate file entirely?

uNetworkingAB:
Can't you put it in WebSocketContextData.h?

It is used by App. and WebSocketContext.h

both of those include (directly or indirectly) WebSocketContextData.h

PR #1557, opened 2023-01-19
Fix reported content length for HttpResponse::endWithoutBody

atimin:
Hey! 
I need to report a real size of a body in the HEAD response. It doesn't work without this patch.  
Thanks.


atimin:
Hey @uNetworkingAB,  could you give your feedback on my changes? No pressure at all, I just need to understand how to proceed.

PR #1579, opened 2023-02-24
Add uws_loop_defer function to uWebsockets CAPI

GenrikhFetischev:
Hello! 
I am working on Rust bindings to uWebSockets, and as far as I understand, defer is the only way to use the library in multiple threads (which is necessary in my case). 
So, this pull request adds a function `uws_loop_defer` to the CAPI.
Thank you for considering that PR.

### DIFF
--- a/capi/libuwebsockets.cpp
+++ b/capi/libuwebsockets.cpp
@@ -23,6 +23,7 @@
 #include "libuwebsockets.h"
 #include <string_view>
 #include "App.h"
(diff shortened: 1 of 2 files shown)


GenrikhFetischev:
Hi @uNetworkingAB!
Could you please clarify if there is any hope that my PR can be merged? If there is something that prevents that PR from considering and that is in my power to fix - please, let me know.

PR #1584, opened 2023-03-03
sprintf -> snprintf to silence compiler warnings about unsafe string …

cmorganBE:
…functions

Compilers may warn about the use of unsafe string functions and users may question the use of sprintf(). Switching to snprintf removes the question about whether unsafe string functions are being used.

Note that the two cases that use buffer_size() were safe as buffer_size() was used during the allocation of the output buffer. Two other uses of sprintf() in the examples may not have been safe in all cases.

### DIFF
--- a/capi/examples/Broadcast.c
+++ b/capi/examples/Broadcast.c
@@ -115,9 +115,9 @@ void on_timer_interval(void* data){
 
     int64_t millis = ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
 
-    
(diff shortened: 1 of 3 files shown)


cmorganBE:
Any variation of this PR of any interest?

uNetworkingAB:
It doesn't really fix anything, it just adds length checks that already cannot be exhausted

cmorganBE:
k. Thoughts on adding pragmas to silence compiler warnings then? I think is osx that was popping up those errors.

PR #1599, opened 2023-04-23
 Dellet my work waht same look for thma sami waht we can do    Revert "exposed loop defer function from capi (#1579)"

Porno77:
This reverts commit 09d2795852bf3217e944ceaea8f3fddf538ada40.


uNetworkingAB:
Mr. porno, I can't understand what you're saying

PR #1625, opened 2023-07-06
Support gcc7 - remove charconv

oscarthorn:
Hi!

Not sure which versions of the c++ spec you want to support or which versions of various compilers. But I noticed that uWS is not usable with gcc7. But the only reason for this (as far as I can tell) is that the charconv header is used, which is not implemented by gcc until gcc8. This PR removes the need for that header.

I could also do a marco if you want those that have charconv to use it and only have this change in other cases. Charconv of course is more efficient.

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -19,7 +19,6 @@
 #define UWS_APP_H
 
 #include <string>
-#include <charconv>
 #include <string_view>
 
 namespace uWS {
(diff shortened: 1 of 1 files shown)


uNetworkingAB:
Target is C++17 and std::from_chars is C++17 so nothing to be done

oscarthorn:
Okej, makes sense, thanks anyway!

uNetworkingAB:
Ålrajt

PR #1639, opened 2023-08-12
Added basic HTTP file server example

mdrobot7:
One issue I had while developing another project was getting the HTTP part of uWS up and running to serve the basic HTML and assets for my webpage, so I thought I'd repackage my code and post it as an example. Other examples (HttpServer.cpp, uWS.js's VideoStreamer.js) didn't work for me. HttpServer didn't compile, and VideoStreamer.js uses flags in the HttpResponse object that aren't present in the C++ library, like `HttpResponse.id`.

This is a synchronous HTTP file streamer, so it's not the best for performance, but it should get new users on track to make something better (like the HttpServer example). I've tested it pretty extensively with several large images, and it has held up.


mdrobot7:
Any particular reason for this? There should really be an example of how to use this basic functionality in the repo, both in my experience and talking to others that used this lib it's not the easiest to set up the first time

uNetworkingAB:
I don't like the example. It is not how you're supposed to use the lib and so cannot be part of its examples.

Issue #1659, opened 2023-09-18
Max request header size?


atkaper:
Hi there,

We are running into an issue where we go over the max total request header size of 4K. Can you please either change it from 4K to 8K, or make it configurable?

https://github.com/uNetworking/uWebSockets/blob/f1b10ed221b36c0d8848998e3e939823f66496eb/src/HttpParser.h#L175

Change From:
```
const size_t MAX_FALLBACK_SIZE = 1024 * 4;
```
To:
```
const size_t MAX_FALLBACK_SIZE = 1024 * 8;
```

Note: there is no real RFC I think on what the proper max size is, but it seems some http servers do honor the 8K as well. See for example this old discussion: https://stackoverflow.com/questions/1097651/is-there-a-practical-http-header-length-limit
It mentions "So to conclude: To be accepted by all web servers above, a request’s request line plus header fields should not exceed 8190 Bytes.".

Thanks!

atkaper:
See also: https://github.com/uNetworking/uWebSockets/pull/1660

atkaper:
As additional information; I think more than 4K is big, but... we have no control over this size where we are in the development chain. The infrastructure adds multiple big load-balancer cookies, and we have an authorization token with a JWT user context in them. We can not make the request size smaller.

gillsoftab:
> As additional information; I think more than 4K is big, but... we have no control over this size where we are in the development chain. The infrastructure adds multiple big load-balancer cookies, and we have an authorization token with a JWT user context in them. We can not make the request size smaller.

[Read](https://github.com/uNetworking/uWebSockets/issues/1658#issuecomment-1721244099)

atkaper:
Hi @gillsoftab, thanks. But my question remains. The 4k is an arbitrary limit, and we can not make our code go under the 4k. I see no reason why we could not go to the more modern value of 8K, which is supported by many servers. None of the other components in our landscape have an issue with our headers.

gillsoftab:
> Hi @gillsoftab, thanks. But my question remains. The 4k is an arbitrary limit, and we can not make our code go under the 4k. I see no reason why we could not go to the more modern value of 8K, which is supported by many servers. None of the other components in our landscape have an issue with our headers.

If you ask us, we would have forced max header length to 2k or even 1k. If larger max header size is needed it's an indication the implementer(s) did something wrong or constructed a suboptimal solution. It might be a good idea for them to talk to the yellow rubber duck until progress is made.

[...]

PR #1661, opened 2023-09-19
Issues 1659

atkaper:
Possible suggestion to solve https://github.com/uNetworking/uWebSockets/issues/1659
Might need to be made more in line with the coding standards, but should work as-is also.

### DIFF
--- a/src/HttpParser.h
+++ b/src/HttpParser.h
@@ -172,7 +172,7 @@ private:
     /* This guy really has only 30 bits since we reserve two highest bits to chunked encoding parsing state */
     unsigned int remainingStreamingBytes = 0;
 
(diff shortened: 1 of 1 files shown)


uNetworkingAB:
I'm generally not a fan of dynamic configs but this only happens on slow path so could be reasonable.

Can you add a prefix like UWS_

atkaper:
@uNetworkingAB renamed it as you suggested.

uNetworkingAB:
https://github.com/uNetworking/uWebSockets/commit/75a525655fc865db4d3175cdcefa5a74a825e4df

atkaper:
UWS_HTTP_MAX_HEADERS_SIZE already in master version, so closing pr.

PR #1697, opened 2024-01-17
Feature / Adopt externally accepted sockets into uWebSockets.

markmaker:
# Description

This allows for sockets that were obtained from an external `accept(2)` to be adopted into uWebSockets. 

Such sockets may come from another library, another external event-loop, or from a forked-server pattern, like [inetd](https://en.wikipedia.org/wiki/Inetd), [systemd-socket-activate](https://www.freedesktop.org/software/systemd/man/latest/systemd-socket-activate.html) etc.

This PR is based on https://github.com/uNetworking/uSockets/pull/214 for uSockets. 

Note: I'm unsure how Github handles PRs with submodules, hope I've done it right.

# Justification

A forked process serving a single websocket session can increase safety, as each user is isolated in a separate process. In addition it can do impersonation via [setuid(2)](https://www.man7.org/linux/man-pages/man2/setuid.2.html) to restrict access on the host to that of an authenticated user. The design can also interface legacy "single-user" back-end software. It can increase robustness, limiting the effects of crashes to one session.

This is similar to what libwebsockets does in its "socket adoption helpers":

https://libwebsockets.org/lws-api-doc-main/html/group__sock-adopt.html

# Instructions for use

The following is simplified code (no error checking) to show the forking pattern:

  ```C++
     // somewhere in a parent process, we listen and accept 
     listener = socket(PF_INET, SOCK_STREAM, 0);
      
      bzero(&srv_addr, sizeof(srv_addr));
      srv_addr.sin_family = AF_INET;
      srv_addr.sin_port = htons(port);
      srv_addr.sin_addr.s_addr = htonl(INADDR_ANY);
      
      bind(listener,
                (const struct sockaddr *)&srv_addr,
                sizeof(srv_addr));
    
      listen(listener, 10);
    
      for (;;) {
        int websocket = accept4(listener, NULL, NULL, SOCK_CLOEXEC | SOCK_NONBLOCK);
        // needs to be non-blocking for sure 
        fcntl(websocket, F_SETFL, fcntl(websocket, F_GETFL, 0) | O_NONBLOCK);
    
        switch (fork()) {
        case -1:                        // Error
              close(websocket);     
              break;                      
    
        case 0:                         // Child 
              close(listener);         //  does not need the listener
              
              startApp(websocket);
        
              _exit(EXIT_SUCCESS);
    
        default:                       // Parent
              close(websocket);   // does not need the websocket 
              break
        }
      }
```

then `startApp()` would look something like this:

  ```C++

void startApp(int fd) {
    uWS::SSLApp({
         .cert_file_name = "cert.pem",
         .key_file_name = "key.pem"
      
    }).ws<UserData>("/*", {
  
          .message = [](auto *ws, std::string_view message, uWS::OpCode opCode) {
              ws->send(message, opCode);
          }
          .close = [&](auto *ws, int code, std::string_view message) {
               // this will be the last of it
               uWS::Loop::get()->defer([&](){
                   app.close();
                });
          },
    }).adoptSocket(fd)
    // run this one websocket session to the end
    .run();
}
```

The `adoptSocket()` can be combined with the usual `listen()`, so this would just be the first socket that launches the server process, such as in a [inetd](https://en.wikipedia.org/wiki/Inetd), [systemd-socket-activate](https://www.freedesktop.org/software/systemd/man/latest/systemd-socket-activate.html) pattern.

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -575,6 +575,12 @@ public:
         return std::move(*this);
     }
 
+    /* adopt an externally accepted socket */
+    TemplatedApp &&adoptSocket(LIBUS_SOCKET_DESCRIPTOR accepted_fd) {
+        httpContext->adoptAcceptedSocket(accepted_fd);
+        return std::move(*this);
+    }
+
     TemplatedApp &&run() {
         uWS::run();
         return std::move(*this);
--- a/src/HttpContext.h
+++ b/src/HttpContext.h
@@ -482,6 +482,11 @@ public:
     us_listen_socket_t *listen(const char *path, int options) {
         return us_socket_context_listen_unix(SSL, getSocketContext(), path, options, sizeof(HttpResponseData<SSL>));
     }
+
+    /* Adopt an externally accepted socket into this HttpContext */
+    us_socket_t *adoptAcceptedSocket(LIBUS_SOCKET_DESCRIPTOR accepted_fd) {
+        return us_adopt_accepted_socket(SSL, getSocketContext(), accepted_fd, sizeof(HttpResponseData<SSL>), 0, 0);
+    }
 };
 
 }


ddevienne:
Not an expert, but another advantage of this model is increased concurrency, it seems to me.

uWebSockets is single-threaded by design, and some servers (like my own) must interface with
_slow_ resources (like Databases) that limit concurrent processing for other _web_ sockets of the server.

With an opt-in forking model on accepting a connection, both safety and concurrency are increased.
So while I didn't look at the PRs' code, I strongly support the concept and feature, FWIW.

uNetworkingAB:
> uWebSockets is single-threaded by design, and some servers (like my own) must interface with
slow resources (like Databases) that limit concurrent processing for other web sockets of the server.

You need to go find a DB connector that was written after 1983, not before. A slow DB query that takes 2 minutes to complete does not, in any way, hinder a single-threaded application from progressing otherwise. If it does, then you haven't written a proper single-threaded application. Strongly recommend looking into how Node.js works.

ddevienne:
> You need to go find a DB connector that was written after 1983, not before.

I take your point Alex. And I'm not talking minutes mind you, even 100ms is an eternity for a server to be blocked serving a single request/message. And the time is not necessarily waiting on IO either, it can be compute-bound too. Thus the barb on not using async-io is not entirely called for :)

But this is kinda OT, and I hope I didn't detract from this PR, which has merit IMHO, if only security wise, beside my point on concurrency across processes, via the fork-mode this PR enables (if I understood correctly).

uNetworkingAB:
> even 100ms is an eternity for a server to be blocked serving a single request/message

You aren't supposed to be blocked. That's the whole point of 1983's non-blocking sockets. That's the whole point of Node.js and similar (uws) projects that use non-blocking IO.

> it can be compute-bound too

If you are compute bound, you aren't going to magically summon new computing resources just by spawning a few processes. Spawning processes will take away CPU-time from your constant budget, not add. You are adding more overhead to be processed by finite hardware, reducing available CPU-time for your actual problem.

uNetworkingAB:
If you have 4 CPU-cores you spawn 4 instances of uWS; 4 event-loops on 4 threads running on the 4 CPU-cores. This is how you make things perform optimally, and this is how posted benchmarks have reached top results.

As an objective scientist, I welcome any solution that can beat the results on specified hardware.

ddevienne:
> If you have 4 CPU-cores you spawn 4 instances of uWS; 4 event-loops on 4 threads running on the 4 CPU-cores.

And you bind all of them to the same port with SO_REUSE? (isn't that Linux only?)
And the OS magically spreads the connections across the different instances (each running on its _own_ core)?

But once a client socket is bound to an instance thus core, it can't leverage the other cores.
So at worse, if most of the active (WebSocket) connections are bound to the same core, aren't the other 3 under-utilized?
I'm not being snarky or anything like that. Genuinely curious.

markmaker:
Can we agree that this discussion is at least somewhat OT? :smile:

I did mention interfacing "legacy" and "single user" applications. These might in-deed run single threaded and synchronously when handling the requests/messages as the "back-end" function. But how to deal with that problem, is an entirely different question! 

Can we agree that this _different question_ has nothing _per se_ to do with adopting sockets, or the usefulness of this PR?

@ddevienne, if I understand correctly, having a (legacy) synchronous/CPU-bound back-end simply _forces you_ to spin up a separate thread for the `uWS::App` (or vice versa for the back-end). Then pass requests/messages from uWS to the synchronous thread using a (thread safe) task queue, much like `uWS::Loop::defer()` does. In fact, I plan to experiment with using a separate `uWS::Loop` on the synchronous thread, just for running deferred tasks.

Under these assumptions, each such synchronous task will then run _serialized_ on the synchronous thread, and not hinder uWS from working _in parallel_ and _asynchronously_ on its own thread. Once the synchronous task has produced (part of) a response, it in turn must defer back to the uWS thread, using `uWS::Loop::defer()`. Inside each such defer call (e.g. inside a lambda), it can use `res->send()` and/or one final `res->end()` for http handlers. Or any number (including zero) `ws->send()` for websockets. For websockets it can defer-send regardless of incoming messages, perhaps from synchronous tasks that run on timers, or other triggers, at any point in time during the life time of the websocket.

If these assumptions are true, and provided that we have threads in C/C++, I would not per se say that the serialized synchronous task design is inferior. In fact, unless I'm mistaken, Node.js does the same thing, using [workers](https://nodejs.org/api/worker_threads.html). When you have CPU bound tasks, an asynchronous design is simply a bad choice. If you have heavy database transactions with [ACID](https://en.wikipedia.org/wiki/ACID) guarantees, [serialization](https://en.wikipedia.org/wiki/Database_transaction_schedule#Serializable) of tasks by user/session is essential, as otherwise you would very likely just block or worse [_deadlock_](https://en.wikipedia.org/wiki/Deadlock) yourself.

Note, when I said "synchronous tasks" it does not have to be a "legacy" single threaded design. It could be a modern thread pool design, but still (for good reasons) _synchronous_ inside the individual tasks. 

Well, that's my 2 cents :sweat_smile: 

I hope this will not derail the PR...

uNetworkingAB:
> But once a client socket is bound to an instance thus core, it can't leverage the other cores.
So at worse, if most of the active (WebSocket) connections are bound to the same core, aren't the other 3 under-utilized?

It's a theoretical scenario that won't happen in practice. In practice, the probability for 1 CPU-core to get the new connection is directly inversely proportional to its load. Meaning - CPU-cores that are less loaded are more likely to accept the socket first, thus spreading the load. And if you use the property "maxLifetime" you can set a maximum connection time of 30 minutes, forcing a certain degree of reconnections and thus rebalancing. But most websocket connections are held for short to medium timeframes anyways.

Even if we consider something that tries to have perfect CPU-balancing; Golang - you still have the problem that, even with 100% perfect CPU-load balancing you still have the same problem, but at a higher abstraction level, on instance level. For example, even if an instance has perfect internal load balancing, all instances in your cluster must be load balanced in some discrete (non-divisible) way, either way.

In practice, minimal inter-thread communication and maximal isolation is the key to performance. Solutions that try to maximize inter-thread communication don't even know the limitations of for instance OpenSSL, which performs really bad if you share the SSL_CTX across threads, since it then requires locking for every, fucking, operation. While instead, holding 1 SSL_CTX isolated to one thread, requires no locking at all.

SSL_CTX sharing is one of the reasons this kind of FD importing was removed in early uws 0.14 since for it to work as people wanted, one would need to make OpenSSL use locks everywhere and it was quite complex (OpenSSL is not even officially thread-safe, their own documentation is very vague on that).

markmaker:
> SSL_CTX sharing is one of the reasons this kind of FD importing was removed in early uws 0.14 since for it to work as people wanted, one would need to make OpenSSL use locks everywhere and it was quite complex (OpenSSL is not even officially thread-safe, their own documentation is very vague on that).

Note that this PR does not change a thing about that. The socket is adopted by uWS which means it now _belongs_ to it and will henceforth be exclusively used by its thread/loop. When the adopt happens, the SSL_CTX is not even assigned yet.

I really hope you look at the actually proposed code, not at this unrelated discussion.

uNetworkingAB:
I can see this being merged in some form. It could be used for explicit load balancing, if nothing else. You could explicitly distribute sockets as you wish rather than use the default built in.

But I need to test this and go over it thoroughly and I can't do that right now

uNetworkingAB:
Hmm, maybe I remember the "transfer" feature we had in 0.14. The problem was that you could transfer sockets between threads by exporting their FD and SSL and then import it, but that was really complex.

If this feature only is importing it from the start, and then never exporting it, that could be a good feature

PR #1792, opened 2024-10-09
add buffer.shrink_to_fit() to reduce memory

gubaojian:
add buffer.shrink_to_fit() to reduce memory when clear buffer.


uNetworkingAB:
This whole class is in need of a rewrite. It's not just the clear, it's also the erase.

PR #1794, opened 2024-10-14
Added a Chinese version of the README file to the project

hufei365:
Added a Chinese version of the README file to the project, which is now ready to be merged into the main branch. 

This addition aims to enhance accessibility and provide a more inclusive experience for Chinese-speaking users who wish to contribute to or utilize the project. 

The translated document maintains the integrity and style of the original content, ensuring that all necessary information is conveyed accurately and effectively. Looking forward to the community's feedback on this update.


uNetworkingAB:
The problem is that I don't know Chinese and even if this would be correct, all documentation and FAQ is still in english so at best this is cosmetical. Btw, is the suffix _ZH automatically picked up?

PR #1866, opened 2025-06-23
use precise mask, which is better than unprecise to avoid effect remain data

gubaojian:
use precise mask, which is better than unprecise to avoid effect remain data


uNetworkingAB:
Feel free to post any security bugs as issues in the tracker. This PR does not fix anything.

PR #1867, opened 2025-06-23
use precise mask to fix bug for part parse lead frame message error 

gubaojian:
use precise mask to fix bug for part parse lead frame message error.  
here is the test code, you can use the code  and the data in zip file to Reproduce the bug :
[one_frame.dat.zip](https://github.com/user-attachments/files/20857351/one_frame.dat.zip)

`

    int main() {
       std::string framedata =   FileUtils::readFile("/Users/baojian/code/ServerTechTest/jwt/cplus/MaskTest/CopyTest/one_frame.dat", true);
    
    std::string two;
    two.append(framedata);
    two.append(framedata);

    uWS::WebSocketProtocol<true, Impl> protocol;
    uWS::WebSocketState<true> state;
    
    char* src = (char*)framedata.data();
    
    
    //part parse
    protocol.consume(src, 20, &state, nullptr);
    protocol.consume(src + 20, 20, &state, nullptr);
    protocol.consume(src + 40, framedata.size() - 40, &state, nullptr);
    
    
    //parse two message
    protocol.consume(two.data(), two.size(), &state, nullptr);

    }

`

### DIFF
--- a/src/WebSocketProtocol.h
+++ b/src/WebSocketProtocol.h
@@ -295,6 +295,24 @@ protected:
         }
     }
 
+   static inline void unmaskPrecise8(char *src, uint64_t maskInt, char* mask, unsigned int length) {
+        size_t lengthu64 = length/8;
+        uint64_t* u64I = (uint64_t*)src;
+        uint64_t* u64O = (uint64_t*)src;
+        for(size_t m=0; m<lengthu64; m++){
+            u64O[m] = u64I[m] ^ maskInt;
+        }
+        size_t remain = length % 8;
+        if (remain > 0) {
+            size_t roffset =  length - remain;
+            char* rI = (src + roffset);
+            char* rO = (src + roffset);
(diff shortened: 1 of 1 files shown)


gubaojian:
`
 unmaskImprecise8<0>(src + MESSAGE_HEADER, mask, length); 

`
above the line code may lead to overflow visit of the src buffer.

uNetworkingAB:
std::string two;
two.append(framedata);
two.append(framedata);


You need to pre-pad, post-pad your buffer with at least 32 bytes. All parsers in uWS build on relaxed boundary cehcks due to how all buffers are pre and post padded.

gubaojian:
> std::string two; two.append(framedata); two.append(framedata);
> 
> You need to pre-pad, post-pad your buffer with at least 32 bytes. All parsers in uWS build on relaxed boundary cehcks due to how all buffers are pre and post padded.

thanks for your reply, seems it works well

PR #1872, opened 2025-06-27
Provide access to existing parameter names

sgoth:
This can be useful in higher level abstractions that outlive HttpRequest.

If you'd prefer a different solution like making the currentParameterOffsets accessible and let the user do the copy please let me know.


uNetworkingAB:
This is the wosts code I've seen in my life. A getter with memory allocations my god.

PR #1876, opened 2025-07-04
 only drop text and binary message when pressure, control message such as ping pong and close message should not be drop

gubaojian:
code in uwebsockets such send ping or send pong or send close, not check the drop status. which will lead to control message lose when backpressure.  the pr will fix this and only drop text or binary message when backpressure

### DIFF
--- a/src/WebSocket.h
+++ b/src/WebSocket.h
@@ -115,20 +115,24 @@ public:
             (us_socket_context_t *) us_socket_context(SSL, (us_socket_t *) this)
         );
 
-        /* Skip sending and report success if we are over the limit of maxBackpressure */
-        if (webSocketContextData->maxBackpressure && webSocketContextData->maxBackpressure < getBufferedAmount()) {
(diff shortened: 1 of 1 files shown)


uNetworkingAB:
Why would control messages be exempt from backpressure logic? Control messages are not critically important, most of them are simple pings or pongs or close frames.

If you've reached the maxBackpressure, that's it. It will drop.

gubaojian:
> Why would control messages be exempt from backpressure logic? Control messages are not critically important, most of them are simple pings or pongs or close frames.
> 
> If you've reached the maxBackpressure, that's it. It will drop.
some ws client use ping pong to keep alive.  client send ping, server return pong. 
if client send ping, not receive pong, it will assume connection lost. when server handle lot of message in backpressure, connection is ok, but pong be dropped. client not not receive pong, client will assume connection be dead. connection will be close by client. this behavior is not right.

PR #1897, opened 2025-11-20
Update HttpResponse with beginWrite

notaphplover:
### Changed
- Updated `HttpResponse` with `beginWrite`.

### Context
See #1895


uNetworkingAB:
This is a good idea, I have missed it. I will probably add this without the tests (the tests are platform dependent and not really deterministic).

notaphplover:
Sure, I'll remove them in a commit. If you decide to go with a separate pr instead of this one, could I ask you to co author me in the commit message? This is probably an act of vanity but figuring as a contributor to your repo would be almost a dream come true.

PR #1921, opened 2026-04-02
adoptSocket: add optional ip parameter

ValoChet:
Related to uWS.js issue [1257](https://github.com/uNetworking/uWebSockets.js/issues/1257).
Depends on uSockets PR [251](https://github.com/uNetworking/uSockets/pull/251).

Update `adoptSocket()` to accept an optional IP parameter.
Update LocalCluster and `addChildApp()` to retrieve the IP on pre_open and pass it to `adoptSocket()`.
Allowing the IP to be correctly set in the uSockets open handler for the adopted socket.

If someone accepts external FD, needs the IP on the request and uses the UWS_REMOTE_ADDRESS_USERSPACE cache, they have to pass the IP to adoptSocket().

### DIFF
--- a/src/App.h
+++ b/src/App.h
@@ -593,7 +593,7 @@ public:
     }
 
     /* Register event handler for accepted FD. Can be used together with adoptSocket. */
-    TemplatedApp &&preOpen(LIBUS_SOCKET_DESCRIPTOR (*handler)(struct us_socket_context_t *, LIBUS_SOCKET_DESCRIPTOR)) {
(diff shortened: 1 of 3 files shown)


src/App.h
```diff
@@ -634,9 +634,9 @@ struct TemplatedApp {
             //std::cout << "Loop is " << receivingApp->getLoop() << std::endl;
 
 
-            receivingApp->getLoop()->defer([fd, receivingApp]() {
+            receivingApp->getLoop()->defer([fd, ipStore = std::vector(ip, ip + ip_length), receivingApp]() {
```
uNetworkingAB: You don't want std::vector for 16 bytes, use std::array<17>

uNetworkingAB: Or std::string (it's short string optimized up to 32 bytes)

ValoChet: Ok, changed it to std::string.

Issue #884, opened 2019-04-10
`tryWrite` method?


aadityataparia:
Is there a `tryWrite` method like `tryEnd` or can there be?

difference with `tryEnd` being it won't require size argument when writing, but will give backpressure handling capability.

Example usage:

```js
    readStream.on('data', (buffer) => {
      const chunk = buffer.buffer.slice(buffer.byteOffset, buffer.byteOffset + buffer.byteLength),
        lastOffset = res.getWriteOffset();

      // First try
      const ok = res.tryWrite(chunk);

      if (!ok) {
        // pause because backpressure
        readStream.pause();

        // Save unsent chunk for later
        res.ab = chunk;
        res.abOffset = lastOffset;

        // Register async handlers for drainage
        res.onWritable((offset) => {
          const ok = res.tryWrite(res.ab.slice(offset - res.abOffset));
          if (ok) {
            readStream.resume();
          }
          return ok;
       });
      }
    });
    readStream.on('end', () => res.end());
```

or are there are alternatives?

PR #1922, opened 2026-04-16
Optimize chunked write path and add tryWrite

GetThatCookie:
Attempt to improve chunked streaming by integrating tryWrite and tightening write

@uNetworkingAB I wanted to try a small uWS-side improvement for the chunked
/ no-content-length HTTP response path discussed around:
- uNetworking/uWebSockets#884
- uNetworking/uWebSockets.js#1249

My intent here was not to present this as a perfect or final answer, but to
test whether integrating `tryWrite` into this area could also let me tighten
the existing `write` path with a small and contained change surface.

What made this seem worth trying:
- I noticed the code still explicitly called out `tryWrite` as not yet
  integrated in this path
- the behavior being discussed in the linked issues pointed more toward the
  underlying no-content-length streaming path than toward any higher-level
  framework overhead
- that made this part of uWS look like a good candidate for a focused
  implementation attempt

What I implemented:
- add `HttpResponse::tryWrite(std::string_view)` for chunked responses
- move chunked `write()` onto a shared internal write core
- route chunked `end()` through that same core so framing, timeout handling,
  completion, and close behavior stay aligned
- keep continuation state so partial optional writes can resume from the
  remaining body suffix via `getWriteOffset()`

Why I thought this was the right experiment:
- it gives unknown-size chunked streaming a primitive analogous in spirit to
  `tryEnd`
- it lets `write`, `tryWrite`, and terminating chunked `end` share one
  implementation instead of evolving independently
- it keeps the attempt focused on uWS itself and avoids a broader
  architectural change

Local benchmark notes:
- on one local async no-content-length streaming benchmark, the `write` path
  moved from about `1073 req/s` before this change to about `3186 req/s`
  after it
- on the resulting implementation, separate local `wrk` checks for the new
  chunked streaming behavior showed:
  - `GET /trywrite` with `128 KiB`: `30882.97 req/s`, `3.77 GB/s`, `p99 3.74 ms`
  - `GET /trywrite-end` with `256 KiB`: `30500.77 req/s`, `7.45 GB/s`, `p99 3.92 ms`

These are only local workload-specific numbers, but they were encouraging
enough for me to keep the change.

Technical notes:
- the core change is that chunked body emission now goes through one internal
  helper instead of keeping separate behavior between `write()` and chunked
  `end()`
- that helper is responsible for:
  - emitting the chunk header once per chunk
  - writing the payload with either optional or non-optional semantics
  - updating the HTTP write offset
  - deciding whether the chunk is complete or still pending
  - emitting the terminating `0\r\n\r\n` when chunked `end()` completes
- for optional writes, I keep a small continuation bit in response state so
  the code can distinguish between:
  - starting a fresh chunk, where a new chunk header must be emitted
  - continuing a previously partial chunk, where only the remaining payload
    suffix should be written
- the continuation logic is intentionally tied to `getWriteOffset()`, so
  `onWritable` can resume from the remaining body suffix without needing a
  second chunk-framing path
- this also lets `tryWrite(...) -> false -> onWritable(...) -> tryWrite(rest)`
  and `tryWrite(...) -> false -> end(rest)` follow one consistent model

Validation:
- I extended the existing `Crc32` / `tests/smoke.mjs` flow to cover
  `write`, `tryWrite`, and `tryWrite + end`
- the affected local unit tests, smoke tests, and HTTP compliance checks
  passed in the validated environment

### DIFF
--- a/examples/ChunkedResponse.cpp
+++ b/examples/ChunkedResponse.cpp
@@ -1,0 +1,64 @@
+#include "App.h"
+
+/* This example demonstrates a large chunked response streamed with tryWrite. */
+
+#include <cstdint>
+#include <memory>
+#include <string>
+
+namespace {
+
+const std::string payload(16 * 1024 * 1024, 'x');
+
+struct ResponseState {
+    uintmax_t baseOffset = 0;
+    bool aborted = false;
+};
+
+template <bool SSL>
+bool tryWriteLoop(uWS::HttpResponse<SSL> *res, ResponseState *state) {
+    if (state->aborted) {
+        return true;
+    }
+
+    uintmax_t sent = res->getWriteOffset() - state->baseOffset;
+    std::string_view remaining = payload;
+    remaining.remove_prefix((size_t) sent);
+    if (res->tryWrite(remaining)) {
+        res->end();
+        return true;
+    }
+
+    return false;
+}
+
+}
+
+int main() {
+
+    uWS::SSLApp({
+      .key_file_name = "misc/key.pem",
+      .cert_file_name = "misc/cert.pem",
+      .passphrase = "1234"
+    }).get("/*", [](auto *res, auto */*req*/) {
+        auto state = std::make_shared<ResponseState>();
+        state->baseOffset = res->getWriteOffset();
+
+        res->writeHeader("Content-Type", "application/octet-stream");
+        res->onAborted([state]() {
+            state->aborted = true;
+        });
+
+        if (!tryWriteLoop(res, state.get())) {
+            res->onWritable([res, state](uintmax_t) {
+                return tryWriteLoop(res, state.get());
+            });
+        }
+    }).listen(3000, [](auto *listen_socket) {
+        if (listen_socket) {
+            std::cout << "Listening on port " << 3000 << std::endl;
+        }
+    }).run();
+
+    std::cout << "Failed to listen on port 3000" << std::endl;
+}
--- a/examples/ChunkedResponse.cpp
+++ b/examples/ChunkedResponse.cpp
@@ -23,12 +23,7 @@ bool tryWriteLoop(uWS::HttpResponse<SSL> *res, ResponseState *state) {
     uintmax_t sent = res->getWriteOffset();
     std::string_view remaining = payload;
     remaining.remove_prefix((size_t) sent);
-    if (res->tryWrite(remaining)) {
-        res->end();
-        return true;
-    }
-
-    return false;
+    return res->tryWrite(remaining, true);
 }
 
 }
--- a/examples/ChunkedResponse.cpp
+++ b/examples/ChunkedResponse.cpp
@@ -11,7 +11,6 @@ namespace {
 const std::string payload(16 * 1024 * 1024, 'x');
 
 struct ResponseState {
-    uintmax_t baseOffset = 0;
     bool aborted = false;
 };
 
@@ -21,7 +20,7 @@ bool tryWriteLoop(uWS::HttpResponse<SSL> *res, ResponseState *state) {
         return true;
     }
 
-    uintmax_t sent = res->getWriteOffset() - state->baseOffset;
+    uintmax_t sent = res->getWriteOffset();
     std::string_view remaining = payload;
     remaining.remove_prefix((size_t) sent);
     if (res->tryWrite(remaining)) {
@@ -42,7 +41,6 @@ int main() {
       .passphrase = "1234"
     }).get("/*", [](auto *res, auto */*req*/) {
         auto state = std::make_shared<ResponseState>();
-        state->baseOffset = res->getWriteOffset();
 
         res->writeHeader("Content-Type", "application/octet-stream");
         res->onAborted([state]() {
--- a/examples/Crc32.cpp
+++ b/examples/Crc32.cpp
@@ -14,6 +14,75 @@
 #include <sstream>
 #include <cstdint>
 #include <cstddef>
+#include <memory>
+#include <string>
+#include <sys/socket.h>
+
+#include "../uSockets/src/libusockets.h"
+
+namespace {
+
+const std::string writeChunk(1024, 'a');
+const std::string tryWritePayload(128 * 1024, 'x');
+const std::string tryWriteEndPayload(256 * 1024, 'y');
+
+struct WriteState {
+    int remaining = 128;
+    bool aborted = false;
+};
+
+struct TryWriteState {
+    const std::string *payload = nullptr;
+    uintmax_t baseOffset = 0;
+    bool aborted = false;
+};
+
+template <bool SSL>
+void setSmallSendBuffer(uWS::HttpResponse<SSL> *res) {
+#ifdef LIBUS_NO_SSL
+    int fd = (int) (uintptr_t) us_socket_get_native_handle(SSL, (us_socket_t *) res);
+    int sendBuffer = 4096;
+    setsockopt(fd, SOL_SOCKET, SO_SNDBUF, &sendBuffer, sizeof(sendBuffer));
+#else
+    (void) res;
+#endif
+}
+
+template <bool SSL>
+void writeLoop(uWS::HttpResponse<SSL> *res, const std::shared_ptr<WriteState> &state) {
+    if (state->aborted) {
+        return;
+    }
+
+    if (!state->remaining) {
+        res->end();
+        return;
+    }
+
+    state->remaining--;
+    res->write(writeChunk);
+    uWS::Loop::get()->defer([res, state]() {
+        writeLoop(res, state);
+    });
+}
+
+template <bool SSL>
+bool tryWriteLoop(uWS::HttpResponse<SSL> *res, const std::shared_ptr<TryWriteState> &state) {
+    if (state->aborted) {
+        return true;
+    }
+
+    uintmax_t sent = res->getWriteOffset() - state->baseOffset;
+    std::string_view remaining(state->payload->data() + sent, state->payload->size() - (size_t) sent);
+    if (res->tryWrite(remaining)) {
+        res->end();
+        return true;
+    }
+
+    return false;
+}
+
+}
 
 uint32_t crc32(const char *s, size_t n, uint32_t crc = 0xFFFFFFFF) {
 
@@ -31,11 +100,57 @@ uint32_t crc32(const char *s, size_t n, uint32_t crc = 0xFFFFFFFF) {
 }
 
 int main() {
-
     uWS::SSLApp({
       .key_file_name = "misc/key.pem",
       .cert_file_name = "misc/cert.pem",
       .passphrase = "1234"
+    }).get("/write", [](auto *res, auto */*req*/) {
+        auto state = std::make_shared<WriteState>();
+
+        res->onAborted([state]() {
+            state->aborted = true;
+        });
+
+        uWS::Loop::get()->defer([res, state]() {
+            writeLoop(res, state);
+        });
+    }).get("/trywrite", [](auto *res, auto */*req*/) {
+        setSmallSendBuffer(res);
+
+        auto state = std::make_shared<TryWriteState>();
+        state->payload = &tryWritePayload;
+        state->baseOffset = res->getWriteOffset();
+
+        res->onAborted([state]() {
+            state->aborted = true;
+        });
+
+        if (!tryWriteLoop(res, state)) {
+            res->onWritable([res, state](uintmax_t) {
+                return tryWriteLoop(res, state);
+            });
+        }
+    }).get("/trywrite-end", [](auto *res, auto */*req*/) {
+        setSmallSendBuffer(res);
+
+        auto state = std::make_shared<TryWriteState>();
+        state->payload = &tryWriteEndPayload;
+        state->baseOffset = res->getWriteOffset();
+
+        res->onAborted([state]() {
+            state->aborted = true;
+        });
+
+        if (!res->tryWrite(*state->payload)) {
+            res->onWritable([res, state](uintmax_t offset) {
+                uintmax_t sent = offset - state->baseOffset;
+                std::string_view remaining(state->payload->data() + sent, state->payload->size() - (size_t) sent);
+                res->end(remaining);
+                return true;
+            });
+        } else {
+            res->end();
+        }
     }).post("/*", [](auto *res, auto *req) {
 
         /* Display the headers */
--- a/examples/Crc32.cpp
+++ b/examples/Crc32.cpp
@@ -14,75 +14,6 @@
 #include <sstream>
 #include <cstdint>
 #include <cstddef>
-#include <memory>
-#include <string>
-#include <sys/socket.h>
-
-#include "../uSockets/src/libusockets.h"
-
-namespace {
-
-const std::string writeChunk(1024, 'a');
-const std::string tryWritePayload(128 * 1024, 'x');
-const std::string tryWriteEndPayload(256 * 1024, 'y');
-
-struct WriteState {
-    int remaining = 128;
-    bool aborted = false;
-};
-
-struct TryWriteState {
-    const std::string *payload = nullptr;
-    uintmax_t baseOffset = 0;
-    bool aborted = false;
-};
-
-template <bool SSL>
-void setSmallSendBuffer(uWS::HttpResponse<SSL> *res) {
-#ifdef LIBUS_NO_SSL
-    int fd = (int) (uintptr_t) us_socket_get_native_handle(SSL, (us_socket_t *) res);
-    int sendBuffer = 4096;
-    setsockopt(fd, SOL_SOCKET, SO_SNDBUF, &sendBuffer, sizeof(sendBuffer));
-#else
-    (void) res;
-#endif
-}
-
-template <bool SSL>
-void writeLoop(uWS::HttpResponse<SSL> *res, const std::shared_ptr<WriteState> &state) {
-    if (state->aborted) {
-        return;
-    }
-
-    if (!state->remaining) {
-        res->end();
-        return;
-    }
-
-    state->remaining--;
-    res->write(writeChunk);
-    uWS::Loop::get()->defer([res, state]() {
-        writeLoop(res, state);
-    });
-}
-
-template <bool SSL>
-bool tryWriteLoop(uWS::HttpResponse<SSL> *res, const std::shared_ptr<TryWriteState> &state) {
-    if (state->aborted) {
-        return true;
-    }
-
-    uintmax_t sent = res->getWriteOffset() - state->baseOffset;
-    std::string_view remaining(state->payload->data() + sent, state->payload->size() - (size_t) sent);
-    if (res->tryWrite(remaining)) {
-        res->end();
-        return true;
-    }
-
-    return false;
-}
-
-}
 
 uint32_t crc32(const char *s, size_t n, uint32_t crc = 0xFFFFFFFF) {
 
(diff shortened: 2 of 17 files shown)


GetThatCookie:
Please feel free to take, change, simplify, or remove any part of this as you
see fit. I am not attached to this going in exactly as-is; the main goal here
was to provide a concrete implementation attempt around this path.

If only parts of it are worth keeping, that is perfectly fine too.

uNetworkingAB:
Quick scroll through, my gut feeling says it's high quality high probability of being merged but I still need to review this for a long time

GetThatCookie:
Thanks for the quick check. Please take your time.

examples/Crc32.cpp
```diff
@@ -14,6 +14,75 @@
...
+};
+
+struct TryWriteState {
+    const std::string *payload = nullptr;
+    uintmax_t baseOffset = 0;
+    bool aborted = false;
+};
+
+template <bool SSL>
+void setSmallSendBuffer(uWS::HttpResponse<SSL> *res) {
```
uNetworkingAB: Not a fan of this

examples/Crc32.cpp
```diff
@@ -14,6 +14,75 @@
...
+    }
+
+    if (!state->remaining) {
+        res->end();
+        return;
+    }
+
+    state->remaining--;
+    res->write(writeChunk);
+    uWS::Loop::get()->defer([res, state]() {
```
uNetworkingAB: Why defer in single threaded app?

examples/Crc32.cpp
```diff
@@ -31,11 +100,57 @@ uint32_t crc32(const char *s, size_t n, uint32_t crc = 0xFFFFFFFF) {
...
       .cert_file_name = "misc/cert.pem",
       .passphrase = "1234"
+    }).get("/write", [](auto *res, auto */*req*/) {
+        auto state = std::make_shared<WriteState>();
+
+        res->onAborted([state]() {
+            state->aborted = true;
+        });
+
+        uWS::Loop::get()->defer([res, state]() {
```
uNetworkingAB: Again with the defer, why?

examples/Crc32.cpp
```diff
@@ -14,6 +14,75 @@
...
+
+    state->remaining--;
+    res->write(writeChunk);
+    uWS::Loop::get()->defer([res, state]() {
+        writeLoop(res, state);
+    });
+}
+
+template <bool SSL>
+bool tryWriteLoop(uWS::HttpResponse<SSL> *res, const std::shared_ptr<TryWriteState> &state) {
```
uNetworkingAB: This makes senese

GetThatCookie:
Addressed the review points.

Changes:
- moved the smoke-only routes out of `Crc32` and into the dedicated `SmokeTest` binary
- removed the `defer` usage and resumed streaming strictly via `onWritable`
- removed the socket send-buffer tweaking and instead use larger smoke payloads to exercise the `tryWrite` path
- updated smoke/CI wiring to run against `SmokeTest`

GetThatCookie:
I reduced the follow-up scope and tried to align it with the existing repo structure.

The net diff now only contains the tryWrite core in HttpResponse/HttpResponseData plus a separate manual example (`examples/ChunkedResponse.cpp`).

This seemed closer to the existing example-based usage pattern already present in the repo than keeping the previous smoke/CI wiring.

src/HttpResponse.h
```diff
@@ -87,6 +94,72 @@ struct HttpResponse : public AsyncSocket<SSL> {
...
+            httpResponseData->state |= HttpResponseData<SSL>::HTTP_WRITE_CALLED;
+        }
+
+        bool completed = !continuingChunk || data.length();
+        bool failed = false;
+        if (data.length()) {
+            if (!continuingChunk) {
+                char chunkHeader[34];
+                unsigned int chunkHeaderLength = formatChunkHeader((unsigned int) data.length(), chunkHeader);
+                failed = Super::write(chunkHeader, (int) chunkHeaderLength).second;
```
uNetworkingAB: Ok so this thing needs to be optional = false, maybe explicitly set it to false and put a comment about how it must not be optional. Note to self: I can make that edit

src/HttpResponse.h
```diff
@@ -87,6 +94,72 @@ struct HttpResponse : public AsyncSocket<SSL> {
 #endif
     }
 
+    /* Chunked writes can only be resumed by continuing the same body suffix. */
+    std::pair<bool, bool> internalWriteChunk(std::string_view data, bool optional, bool terminate = false) {
+        HttpResponseData<SSL> *httpResponseData = getHttpResponseData();
+        bool continuingChunk = httpResponseData->state & HttpResponseData<SSL>::HTTP_WRITE_CONTINUATION_PENDING;
```
uNetworkingAB: This is whether we are inside a single chunk or not

src/HttpResponse.h
```diff
@@ -87,6 +94,72 @@ struct HttpResponse : public AsyncSocket<SSL> {
...
+        bool failed = false;
+        if (data.length()) {
+            if (!continuingChunk) {
+                char chunkHeader[34];
+                unsigned int chunkHeaderLength = formatChunkHeader((unsigned int) data.length(), chunkHeader);
+                failed = Super::write(chunkHeader, (int) chunkHeaderLength).second;
+            }
+
+            auto writtenFailed = Super::write(data.data(), (int) data.length(), optional);
+            httpResponseData->offset += (uintmax_t) writtenFailed.first;
```
uNetworkingAB: Ok the interface builds on getWriteOffset (what you get in onWritable). I guess this is the same for tryEnd.

examples/ChunkedResponse.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+int main() {
+
+    uWS::SSLApp({
+      .key_file_name = "misc/key.pem",
+      .cert_file_name = "misc/cert.pem",
+      .passphrase = "1234"
+    }).get("/*", [](auto *res, auto */*req*/) {
+        auto state = std::make_shared<ResponseState>();
+        state->baseOffset = res->getWriteOffset();
```
uNetworkingAB: Why isn't baseOffset always 0 for a new request?

src/HttpResponse.h
GetThatCookie: Made it explicitly `false` and added a comment about why it must not be optional

src/HttpResponse.h
```diff
@@ -87,6 +94,72 @@ struct HttpResponse : public AsyncSocket<SSL> {
...
+            writeMark();
+            writeHeader("Transfer-Encoding", "chunked");
+            httpResponseData->state |= HttpResponseData<SSL>::HTTP_WRITE_CALLED;
+        }
+
+        bool completed = !continuingChunk || data.length();
+        bool failed = false;
+        if (data.length()) {
+            if (!continuingChunk) {
+                char chunkHeader[34];
```
uNetworkingAB: Why is this not 12?

src/HttpResponse.h
GetThatCookie: Renamed it to `insideChunk` to reflect that more directly

src/HttpResponse.h
GetThatCookie: Added a comment clarifying that this offset tracks body bytes only, matching `getWriteOffset()` and the offset passed to `onWritable`

examples/ChunkedResponse.cpp
GetThatCookie: You're right, for this example it is always a new response offset, so I removed `baseOffset` and use `getWriteOffset()` directly.

src/HttpResponse.h
```diff
@@ -73,6 +71,15 @@ struct HttpResponse : public AsyncSocket<SSL> {
         Super::write(buf, length);
     }
 
+    unsigned int formatChunkHeader(unsigned int value, char *dst) {
```
uNetworkingAB: Can we name this "formatCRLFAndChunkHeader" since the first CRLF does not actually belong to the chunkHeader

src/HttpResponse.h
GetThatCookie: Good catch. `12` is sufficient here: `\r\n` + 8 hex digits + `\r\n`

src/HttpResponse.h
GetThatCookie: Nothing easier than that, renamed to `formatCRLFAndChunkHeader`

src/AsyncSocketData.h
```diff
@@ -18,13 +18,27 @@
 #ifndef UWS_ASYNCSOCKETDATA_H
```
uNetworkingAB, line 18: Plese don't widen the scope with even more changes - put that in a different PR if so

GetThatCookie: I was just about to leave a comment about this because I saw the story from #1792 and the immediate impact it has on the same write/tryWrite path being tightened here.

I do think it is reasonable to split this into a follow-up PR if you prefer, but I would personally prefer to keep it in this PR since it directly affects the same path. I also would not want to add anything further beyond this.

The reason I mentioned it here is that it directly affects the same path #1922 is optimizing. Even with the chunked `write` / `tryWrite` refactor, the buffered write/drain path still runs through `AsyncSocket::write` and then `BackPressure::{erase,clear,resize}` whenever user-space backpressure is present.

So this is not an unrelated cleanup; it reduces overhead in the same write/drain path that this PR exercises under backpressure.

I revisited the unconditional `shrink_to_fit()` behavior from #1792 and benchmarked a local `BackPressure` rewrite. Against the current behavior, local results were:

- `4 KB` steady drain: `49.42 ns -> 27.82 ns` (`-43.7%`)
- `32 KB` steady drain: `477.51 ns -> 406.07 ns` (`-15.0%`)

Targeted patterns were also faster:
- `erase 16 KB`, then `append 4 KB`: about `-49%`
- many small erases before append: about `-55%`
- `erase 1 KB`, then `resize + 1 KB`: about `-42%`

So the point is not that this speeds up the raw socket write syscall, but that it makes the buffered write path cheaper in exactly the area that #1922 is tightening.

GetThatCookie: If that sounds reasonable to you, I would prefer to keep it here. Otherwise I can split it out into a small follow-up PR.

GetThatCookie: From my side this PR is done now. If you agree with keeping this here, feel free to resolve the conversation.

src/AsyncSocketData.h
```diff
@@ -18,13 +18,27 @@
...
 #define UWS_ASYNCSOCKETDATA_H
 
+#include <cstring>
 #include <string>
 
 namespace uWS {
 
 struct BackPressure {
+    static constexpr size_t TRIM_THRESHOLD = 64 * 1024;
+    static constexpr size_t RETAIN_CAPACITY = 32 * 1024;
```
uNetworkingAB: This is definitely not right, uWS has never used per socket user space buffers like this. We rely on the kernel being ready to take data directly (that's the fast path). If you retain 32kb per socket, that's missing the point of the cork buffer - per socket backpressure should only be when you actually have backpressure, and that's the slow path. If you properly use tryWrite or tryEnd, you'll never end up in the backpressure buffer. So retaining 32kb is just arbitrary and if you get speedups from that, then the app or heuristics in the lib, needs better use of the corkbuffer. Highly dislike this retainment

uNetworkingAB: Allocating 32kb is also very fast in libtcmalloc, jemalloc and the like, as it hits the free list. So retaining such a small but still relatively big memory always, is a big downgrade to memory use

src/AsyncSocketData.h
```diff
@@ -18,13 +18,27 @@
...
+
     std::string buffer;
     unsigned int pendingRemoval = 0;
+
+    void normalize() {
+        size_t length = buffer.length() - pendingRemoval;
+        if (length) {
+            memmove(buffer.data(), buffer.data() + pendingRemoval, length);
```
uNetworkingAB: I shoved this into compilerexplorer and it looks like `std::string(buffer.begin() + pendingRemoval, buffer.end()).swap(buffer);` is a shit storm of bloat in comparison, so that's good (and it is even easier to understand)

src/AsyncSocketData.h
uNetworkingAB: From what I can tell, your tryWrite HTTP changes have no dependency on the changes to AsyncSocketData.h‎ and changes to AsyncSocketData.h‎ have vastly different reach. Definitely put that in a separate PR and for that PR there needs to be thorough benchmarks in memory use and all that. So definitely a separate PR

src/AsyncSocketData.h
uNetworkingAB: Mind you that, once we have tryWrite, any use of write should be seen the same as end is to tryEnd. That is, any app that does not properly use the tryWrite, tryEnd have absolutely no relevance in benchmarks (we aren't optimizing the slow paths)

examples/ChunkedResponse.cpp
```diff
@@ -0,0 +1,62 @@
...
+template <bool SSL>
+bool tryWriteLoop(uWS::HttpResponse<SSL> *res, ResponseState *state) {
+    if (state->aborted) {
+        return true;
+    }
+
+    uintmax_t sent = res->getWriteOffset();
+    std::string_view remaining = payload;
+    remaining.remove_prefix((size_t) sent);
+    if (res->tryWrite(remaining)) {
```
uNetworkingAB: These 2 calls will cause 2 syscalls. If tryWrite takes `bool finalChunk` the 2 chunks can be sent with us_socket_write2

src/HttpResponse.h
```diff
@@ -452,33 +492,23 @@ struct HttpResponse : public AsyncSocket<SSL> {
...
-            httpResponseData->state |= HttpResponseData<SSL>::HTTP_WRITE_CALLED;
-        }
+        return !internalWriteChunk(data, false).second;
+    }
 
-        Super::write("\r\n", 2);
-        writeUnsignedHex((unsigned int) data.length());
-        Super::write("\r\n", 2);
+    /* Try and write one chunk. Continue with the remaining body suffix on onWritable. */
+    bool tryWrite(std::string_view data) {
```
uNetworkingAB: This should be `std::string_view data, bool finalChunk = false` where if finalChunk is set, write should happen with us_socket_write2

uNetworkingAB: That optimization could be a separate PR, though

And that should probably be its own function in AsyncSocket::write2

src/AsyncSocketData.h
GetThatCookie: Makes sense. I dropped the AsyncSocketData changes from this PR.

I agree they have wider reach than the HTTP tryWrite changes and should be evaluated separately, with their own memory/perf discussion, if revisited.

src/AsyncSocketData.h
GetThatCookie: Fair point. I removed that retainment/compaction change from this PR.

That part clearly changes the per-socket backpressure behavior and does not need to be coupled to the HTTP tryWrite work here.

GetThatCookie:
I implemented the `finalChunk` / `write2` idea.

The only subtle part was keeping `getWriteOffset()` body-only in the combined path. This was not an issue in the previous PR shape, since the terminating chunk was still sent separately by `end()`.

With `tryWrite(..., true)`, if the body is fully written but only part of the terminating `\r\n0\r\n\r\n` gets out, that trailer should not turn into another public continuation state.

So the current version does the smallest thing:
- `tryWrite(std::string_view, bool finalChunk = false)`
- non-SSL final chunk uses `AsyncSocket::write2`
- offsets stay body-only
- any partial terminating trailer is completed internally
- `ChunkedResponse` now uses `tryWrite(remaining, true)`

Your suggestion pointed to a cleaner path, and my inner monk would not let me leave it half-done once that became obvious. If you want to keep the previous PR shape and leave this optimization for later, I can revert this part.

examples/ChunkedResponse.cpp
```diff
@@ -23,12 +23,7 @@ bool tryWriteLoop(uWS::HttpResponse<SSL> *res, ResponseState *state) {
     uintmax_t sent = res->getWriteOffset();
     std::string_view remaining = payload;
     remaining.remove_prefix((size_t) sent);
-    if (res->tryWrite(remaining)) {
```
uNetworkingAB, line 26: This interface change is chefs kiss. Now it's really simple to use, and allows best case 1 syscall for the entire response just like tryEnd

uNetworkingAB:
It would be cool to see a new release of ultimate express with the current write feature before I continue review this

nigrosimone:
> It would be cool to see a new release of ultimate express with the current write feature before I continue review this

Released today
https://www.npmjs.com/package/ultimate-express/v/2.1.1

GetThatCookie:
I did one more pass after rebasing this PR onto current master and kept the final adjustments contained to the existing implementation.

What I changed:

- large non-SSL POSIX final writes can now use the existing `write2` / `writev` path;
- small writes and the existing SSL, Windows, and io_uring fallbacks remain unchanged;
- fixed chunk framing after `beginWrite()`;
- kept public write offsets limited to body bytes while completing partial trailers internally;
- made backpressure visible in the call that encounters it;
- addressed the CodeQL unused-variable warning without changing behavior.

The reviewed `tryWrite(data, finalChunk)` interface and `ChunkedResponse` example remain unchanged.

### Local A/B benchmark

Before is the rebased PR before this final pass (`2f0d800`); after is the final candidate. Both builds used the same corrected uSockets source.

Results are medians of five interleaved runs with `wrk -t8 -c64`:

| Payload | Before | After | Delta |
| ---: | ---: | ---: | ---: |
| 8 KiB | 351,111 req/s | 353,480 req/s | +0.67% |
| 16 KiB | 215,248 req/s | 272,712 req/s | +26.70% |
| 64 KiB | 130,634 req/s | 149,300 req/s | +14.29% |
| 128 KiB | 94,082 req/s | 97,952 req/s | +4.11% |
| 1 MiB | 16,329 req/s | 16,354 req/s | +0.15% |

At 128 KiB this reduces three separate header/body/trailer sends to one header send with `MSG_MORE`, followed by one `writev` for the body and terminating chunk. The optimized binary adds only 20 bytes of `.text`.

These are local workload-specific results, but they confirm that the new path produces the intended syscall reduction without regressing small or large responses.

### Required uSockets companion fix

I also found that partial `us_socket_write2` calls need the same failure tracking already used by `us_socket_write`.

The required change is in `uSockets/src/socket.c`, inside `us_socket_write2`, currently at line 167:

```c
if (written != header_length + payload_length) {
    s->context->loop->data.last_write_failed = 1;
    us_poll_change(&s->p, s->context->loop, LIBUS_SOCKET_READABLE | LIBUS_SOCKET_WRITABLE);
}
```

Without `s->context->loop->data.last_write_failed = 1`, a 16 MiB response stalled locally after roughly 3.7 MiB under partial backpressure. With the line added, the response completed normally across 12–15 real `onWritable` callbacks.

I would keep this as a separate one-line uSockets change rather than adding a uSockets submodule update to this PR.

All seven local ASan unit executables, strict warning builds, chunk-framing checks, partial body/trailer continuation, and the 16 MiB backpressure case passed.

src/AsyncSocket.h
```diff
@@ -333,17 +333,81 @@ struct AsyncSocket {
...
+    std::pair<int, bool> uncork(const char *src = nullptr, int length = 0, bool optionally = false, bool hasMore = false) {
         LoopData *loopData = getLoopData();
 
         if (loopData->corkedSocket == this) {
             loopData->corkedSocket = nullptr;
 
             if (loopData->corkOffset) {
                 /* Corked data is already accounted for via its write call */
-                auto [written, failed] = write(loopData->corkBuffer, (int) loopData->corkOffset, false, length);
+                bool failed = write(loopData->corkBuffer, (int) loopData->corkOffset, false, length ? length : (int) hasMore).second;
```
uNetworkingAB: What happened here?

src/AsyncSocket.h
```diff
@@ -333,17 +333,81 @@ struct AsyncSocket {
...
+#ifndef _WIN32
+            /* POSIX write2 maps to writev. Keep small writes in the cork buffer, but flush
+             * an existing prefix before a large write so both buffers can use that path. */
+            if (!asyncSocketData->buffer.length() && loopData->corkedSocket == this &&
+                (unsigned int) length > LoopData::CORK_BUFFER_SIZE - loopData->corkOffset) {
+                uncork(nullptr, 0, false, true);
+            }
+#endif
+
+            if (!asyncSocketData->buffer.length() && loopData->corkedSocket != this) {
```
uNetworkingAB: This needs to be isCorked, right?

src/AsyncSocket.h
GetThatCookie: CodeQL flagged written as unused. I changed this to extract only the failed result; there is no behavioral change here. The hasMore handling was already present before that cleanup.

uNetworkingAB: Yeah but length ? length : (int) hasMore looks like a hack to pass 1 even if the size is not 1

PR #1933, opened 2026-07-30
Avoid crashes when publishing without WebSocket contexts

### LINKED ISSUE #1924: SIGSEGV in app.publish() / app.numSubscribers() without calling App.ws to define a WebSocketContext
I'm getting segfaults when I try to publish to a topic to which no subscription has been issued.

It's a pretty simple repro, basically boils down to (more descriptive repro attached).

```js
import { App } from 'uWebSockets.js';
App().listen(8451, () => App().publish('never', 'hi'));  // exit 139
```

[repro-uws-segfault.js](https://github.com/user-attachments/files/27504357/repro-uws-segfault.js)

GetThatCookie:
Fixes #1924.

Return `false` from `publish` and `publishPrepared`, and `0` from `numSubscribers`, when no `WebSocketContext` has been configured.

This avoids dereferencing a null `TopicTree` without allocating one or registering loop handlers for HTTP-only apps.

A regression check was added to the existing HTTP-only fuzz target.

Tested by:
- Building all examples
- Running the complete ASan unit test suite
- Running a targeted ASan/UBSan regression test

### DIFF
--- a/fuzzing/AsyncEpollHelloWorld.cpp
+++ b/fuzzing/AsyncEpollHelloWorld.cpp
@@ -36,6 +36,14 @@ void test() {
             listen_socket = listenSocket;
         });
 
+        /* Publishing without WebSocket contexts must be a no-op */
+        uWS::PreparedMessage preparedMessage = {};
+        if (app.publish("unused", "hi", uWS::TEXT) ||
+            app.publishPrepared("unused", preparedMessage) ||
+            app.numSubscribers("unused")) {
+            exit(-1);
+        }
+
         app.run();
     }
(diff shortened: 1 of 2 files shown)


uNetworkingAB:
Btw, you come from Node.js right? If so, how did you tackle the NPM change for Git dependencies? You just vendor the binaries?

GetThatCookie:
Not originally, but Node.js has pretty much become my new world. I’m still using npm 10 because newer npm versions have introduced too many issues in my workflows. So I haven’t had to deal with the Git dependency change yet - I just install the pinned Git dependency normally and don’t vendor the binaries.

PR #1935, opened 2026-08-11
feat: add HttpRequest::hasBody()

### LINKED ISSUE #1910: Method to indicate that request has body
Is it possible to add method `hasBody()`, which will return `true` if request has body, in order to understand that we need to read it using `onData`.

Soundcreates:
Closes #1910.

Add `HttpRequest::hasBody()` so request handlers can determine whether the request declares a body before registering an `onData` handler.

The helper recognizes both `Content-Length` and `Transfer-Encoding`, matching the parser's body-selection logic. Tests cover requests with and without a declared body.

Validation:
- `make -C tests default`
- AddressSanitizer enabled by the test target

### DIFF
--- a/src/HttpParser.h
+++ b/src/HttpParser.h
@@ -145,6 +145,11 @@ public:
         return std::string_view(headers->key.data(), headers->key.length());
     }
 
+    /* Returns whether the request declares a body that can be delivered through onData. */
+    bool hasBody() {
+        return getHeader("content-length").length() || getHeader("transfer-encoding").length();
+    }
+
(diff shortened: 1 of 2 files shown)


uNetworkingAB:
This implementation is not good. First off, it adds no value to the lib that you couldn't just get from getHeader yourself. Secondly, it re-searches for headers when the parser already did so (the parser already knows the body size). Also, if you get content-length: 0 then hasBody() will be true even though there is no body. Also, you could make the interface much better by returning the actual size if known.

If you want hasBody(), look at how the onDataV2 is implemented and copy that logic

PR #1944, opened 2026-09-17
Support valueless queries

Dev-next-gen:
While reading `getDecodedQueryValue` I noticed that a statement without an equal sign stops the whole search, but only when its first character happens to match the first character of the key being looked up. So with `?debug&dx=5`, `getQuery("dx")` returns nullptr, while `?xdebug&dx=5` correctly returns `"5"`. Valueless flags like `?debug` or `?download` are common in real URLs, so the result of a lookup ends up depending on the spelling of an unrelated parameter.

The early return came in with c3fa05b, which added the `npos` check so that `substr(npos + 1)` could no longer throw. That check is still there. This change drops the `return` in the `else` branch, so the loop moves on to the next statement the same way it already does for statements whose first character doesn't match.

Following the discussion below, the three cases are now distinct: a missing key gives nullptr, `key=` gives an empty string pointing into the buffer, and a bare `key` gives an empty string whose `data()` is `uWS::UWS_VALUELESS_QUERY`.

I added a case to `tests/Query.cpp`. Before the change it fails with `Assertion failed: uWS::getDecodedQueryValue("dx", ...) == "5"`, and after the change the whole file passes. I also ran it under `-fsanitize=address,undefined` with gcc and it came out clean.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


uNetworkingAB:
Interesting but in that case,

we need to differentiate between "not found" and "found but no = sign" and "found with = sign but no value"

* "found with = sign but no value" should probably just give "" (empty string)
* "not found" should give nullptr
* "found but no = sign" - this one is the tricky. It clearly should not give nullptr because it was found, but should it give "" (emptry string?). Ah, I have a solution! We return empty string, but we return an empty string with a specific address like UWS_VALUELESS_QUERY.

That way, users who don't care about the distinction simply see empty string for both valueless and value="" and those who want to differentiate can check if the empty string == UWS_VALUELESS_QUERY

Dev-next-gen:
Returning an empty string at a fixed address keeps the common case simple and still lets callers tell the two apart, so I implemented it that way in 037a5df. A missing key gives nullptr. `key=` gives an empty view into the buffer, as it already did. A bare `key` now gives an empty view whose `data()` is `uWS::UWS_VALUELESS_QUERY`, which is declared in QueryParser.h as `inline constexpr char UWS_VALUELESS_QUERY[1] = {};`. I put it inside `namespace uWS` next to the parser, so it is easy to rename or move if you want it spelled differently.

I made it an inline variable so that the comparison holds across translation units. I compiled two .cpp files that both include the header with g++ 15.2 at -O2, and both lookups returned the same address as the constant (`tu1=0x5606403c5004 tu2=0x5606403c5004 const=0x5606403c5004`). tests/Query.cpp now checks all three cases on `?debug&empty=&x=1`, and also checks that `debu` and `debugger` are still not found and that a valueless key at the end of the query (`?a=1&flag`) is found. The previous head 5cbdf7f fails the new case (`Assertion valueless == "" && valueless.data() == ... failed`). With 037a5df the file passes under g++ 15.2 with `-fsanitize=address,undefined` and under clang 22.1.7 (llvm-mingw). Note that the behaviour changes for code that checked `getQuery("debug").data() == nullptr` to detect a bare flag, since that lookup is now non-null.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

uNetworkingAB:
Gemini tells me it needs to be

// Guaranteed unique address across all translation units
inline constexpr char UWS_VALUELESS_QUERY[] = "";

not

inline constexpr char UWS_VALUELESS_QUERY[1] = {};

Dev-next-gen:
The string literal reads better and says what the constant is, so I switched to `inline constexpr char UWS_VALUELESS_QUERY[] = "";` in 4cc0d00.

For what it's worth, both spellings declare the same object: `static_assert(std::is_same_v<decltype(UWS_VALUELESS_QUERY), const char[1]>)` holds for each of them under clang 22.1.7, and the single address across translation units comes from `inline`. I re-ran the two-TU check with the new spelling (`tu1=00007FF62FC630BD tu2=00007FF62FC630BD const=00007FF62FC630BD`), and tests/Query.cpp still passes under clang 22.1.7 and under g++ 15.2 with `-fsanitize=address,undefined`.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

uNetworkingAB:
Ah, now it tells me they are the same:

> Yes, both declarations guarantee a unique, identical address across all translation units within the same linked program.

Lol fucking hallucinatory nonsense

uNetworkingAB:
This PR is excellent

